PR #2, opened 2022-06-08
Refactor/crypto

foxkdev:
Refactor crypto to not use dependency injection to include in domains
- Now domain file and folder return name decrypted directly.
- Test fixes with typed


src/externals/crypto/aes.ts
```diff
@@ -103,10 +99,11 @@ export class AesService {
     );
 
     // AES 256 GCM Mode
-    const decipher = crypto.createDecipheriv('aes-256-gcm', key, iv);
+    const decipher: any = crypto.createDecipheriv('aes-256-gcm', key, iv);
```
sg-gs, line 102: Avoid using `any`, crypto has the Decipher type, which is in fact automatically typed. Therefore this `any` could be removed directly.

src/externals/crypto/crypto.ts
```diff
@@ -1,16 +1,14 @@
-import { Injectable, Logger } from '@nestjs/common';
-import { AesService } from './aes.service';
+import { Logger } from '@nestjs/common';
+import { AesService } from './aes';
 import CryptoJS from 'crypto-js';
-import { ConfigService } from '@nestjs/config';
+// import { ConfigService } from '@nestjs/config';
```
sg-gs, line 4: Remove this comment.

src/modules/file/file.domain.ts
```diff
@@ -64,6 +66,10 @@ export class File implements FileAttributes {
   static build(file: FileAttributes): File {
     return new File(file);
   }
+  decryptName(name) {
+    const cryptoService = new CryptoService();
```
sg-gs, line 70: Using a singleton for instancing a crypto service could be a better idea.

src/modules/file/file.service.ts
```diff
@@ -1,14 +1,12 @@
 import { Injectable } from '@nestjs/common';
-import { CryptoService } from '../../externals/crypto/crypto.service';
+// import { CryptoService } from '../../externals/crypto/crypto.service';
```
sg-gs, line 2: Remove this comment

src/modules/file/file.service.ts
```diff
@@ -1,14 +1,12 @@
...
 import { FolderAttributes } from '../folder/folder.domain';
 import { FileAttributes } from './file.domain';
 import { SequelizeFileRepository } from './file.repository';
-
 @Injectable()
 export class FileService {
   constructor(
-    private fileRepository: SequelizeFileRepository,
-    private cryptoService: CryptoService,
+    private fileRepository: SequelizeFileRepository, // private cryptoService: CryptoService,
```
sg-gs, line 9: Remove this comment

sg-gs (changes requested):

src/externals/crypto/aes.ts
```diff
@@ -75,13 +71,13 @@ export class AesService {
 
   decrypt(encdata, folderId) {
     // base64 decoding
-    const bData = Buffer.from(encdata, 'base64');
+    const bData: Buffer = Buffer.from(encdata, 'base64');
```
sg-gs, line 74: These types are trivially inferred by TypeScript, there is no need to add them. You should add them when the result of a function is any, unknown, or you are using concrete implementations but you want to force the usage of the interface contract. Example: 

```
interface FilesRepository {
   findAll(): Promise<...>
}

class MongoDBFilesRepository implements FilesRepository {
   ....
   methodThatIsNotInTheInterface()...
}

class MariaDBFilesRepository implements FilesRepository {
  ....
}

// here you want to force the interface usage, so the code will 
// be compatible instantly if it is replaced by MariaDB in any situation
const repository: FilesRepository = new MongoDBRepository();
repository.methodThatIsNotInTheInterface() // TypeScript won't allow this as it can not ensure that this a MongoDBRepository.

```

src/externals/crypto/aes.ts
```diff
@@ -94,7 +90,7 @@ export class AesService {
     }
 
     // derive key using; 32 byte key length
-    const key = crypto.pbkdf2Sync(
+    const key: crypto.CipherKey = crypto.pbkdf2Sync(
```
sg-gs, line 93: Same here if the method does not give any issue, ofc

src/externals/crypto/crypto.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { CryptoService } from './crypto';
+
+describe('Crypto', () => {
+  describe('check crypto as singleton', () => {
+    it('create 2 instances and are the same', () => {
+      const firstInstance = CryptoService.getInstance();
+      const secondInstance = CryptoService.getInstance();
+      expect(firstInstance === secondInstance).toBe(true);
```
sg-gs, line 8: Is `toBe()` enough to deeply check object equality?

sg-gs (changes requested):

src/externals/crypto/crypto.spec.ts
foxkdev: sure, you can use toBe to validate value directly:
https://jestjs.io/es-ES/docs/expect#tobevalue

sg-gs (approved):

PR #6, opened 2022-06-23
Feature/DS-24 analytics lib

foxkdev:
- Added event when get share by token as ShareLinkView.
- Added listener ShareLinkView to send track to segment analytics.


src/lib/analytics.ts
```diff
@@ -0,0 +1,49 @@
...
+  ShareLinkCopied = 'Share Link Copied',
+}
+
+export default class Analytics {
+  analytics: AnalyticsSegment;
+  logger: Logger;
+  static instance: Analytics;
+  constructor() {
+    this.logger = new Logger();
+    this.analytics = new AnalyticsSegment(process.env.APP_SEGMENT_KEY || 'xxx');
```
sg-gs, line 16: avoid defaults for critical values (often related to environment variables) please

sg-gs (changes requested):

sg-gs (approved):

PR #14, opened 2022-07-05
added encryption to name

foxkdev:


src/lib/binary-converter.ts
```diff
@@ -0,0 +1,25 @@
+export function toBinary(string) {
+  const codeUnits = new Uint16Array(string.length);
```
fabioespinosa, line 2: change name to `convertStringToBinary`

src/lib/binary-converter.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+  const charCodes = new Uint8Array(codeUnits.buffer);
+  let result = '';
+  for (let i = 0; i < charCodes.byteLength; i++) {
+    result += String.fromCharCode(charCodes[i]);
+  }
+  return result;
+}
+
+export function fromBinary(binary) {
```
fabioespinosa, line 14: change name to `getStringFromBinary`

fabioespinosa (changes requested):

src/modules/send/send-link.repository.ts
```diff
@@ -19,6 +19,7 @@ import {
   HasMany,
   Sequelize,
 } from 'sequelize-typescript';
+import { fromBinary, toBinary } from 'src/lib/binary-converter';
```
fabioespinosa, line 22: do we need relative imports here? (no idea actually)

fabioespinosa (approved):
🔥

PR #19, opened 2022-07-07
remove type file from mail listener

foxkdev:


src/externals/notifications/listeners/send-link.listener.ts
```diff
@@ -23,7 +23,7 @@ export class SendLinkListener {
 
     const itemsToMail = items.map((item) => {
       return {
-        name: `${item.name}.${item.type}`,
+        name: `${item.name}`,
```
fabioespinosa, line 26: if type is included in name leave it as `name: item.name` without the ``

fabioespinosa (approved):

PR #45, opened 2022-08-25
[WT-33] Feat: add bridge service

JoanVicens:
Added a service to call to be able to call bridge endpoints

### DIFF
--- a/src/externals/bridge/bridge.module.ts
+++ b/src/externals/bridge/bridge.module.ts
@@ -11,7 +11,7 @@ import { BridgeService } from './bridge.service';
     BridgeService,
     {
       provide: CryptoService,
-      useValue: new CryptoService(),
+      useValue: CryptoService.getInstance(),
     },
   ],
   exports: [BridgeService],
--- a/src/externals/bridge/bridge.module.ts
+++ b/src/externals/bridge/bridge.module.ts
@@ -1,0 +1,19 @@
+import { Module } from '@nestjs/common';
+import { ConfigModule } from '@nestjs/config';
+import { CryptoService } from '../crypto/crypto';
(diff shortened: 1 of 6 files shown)


src/externals/bridge/bridge.service.ts
```diff
@@ -0,0 +1,54 @@
...
+    };
+
+    Logger.log(
+      '[INXT removeFile]: User: %s, Bucket: %s, File: %s',
+      user.bridgeUser,
+      bucket,
+      bucketEntry,
+    );
+
+    await this.httpClient.delete(
```
fabioespinosa, line 49: You can remove async and await and just return the promise here

JoanVicens: The http client is returning a `Promise<AxiosResponse<any, any>>`

JoanVicens:
> is it ready for review? or draft?

It is ready, was checking the `hashSha256` method, it was surrounded by a try catch in drive-server but I think its not necessary

src/externals/bridge/bridge.module.ts
```diff
@@ -0,0 +1,19 @@
...
+import { BridgeService } from './bridge.service';
+
+@Module({
+  imports: [ConfigModule, HttpClientModule],
+  controllers: [],
+  providers: [
+    BridgeService,
+    {
+      provide: CryptoService,
+      useValue: new CryptoService(),
```
fabioespinosa, line 14: Why do you have to do this like this with CryptoService here?

JoanVicens: The `CryptoService` was not made "@Injectable()", instead is a singleton. I could use `CryptoService.getInstance()` too

PR #49, opened 2022-08-29
[WT-30]: Feat/migration to postgres wip

fabioespinosa:
Tested with postgres locally, everything seemed ok

Before deploying this to prod we need to make one last sync between the tables ~3 mins

### DIFF
--- a/src/modules/send/send.usecase.ts
+++ b/src/modules/send/send.usecase.ts
@@ -1,9 +1,13 @@
-import { Injectable, NotFoundException } from '@nestjs/common';
+import {
+  BadRequestException,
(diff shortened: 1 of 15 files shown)


src/modules/send/send.usecase.ts
```diff
@@ -15,6 +19,9 @@ export class SendUseCases {
   ) {}
 
   async getById(id: SendLinkAttributes['id']) {
+    if (!validate(id)) {
+      throw new BadRequestException('id is not in uuid format');
```
fabioespinosa, line 23: @foxkdev I added a badrequest here

PR #50, opened 2022-08-29
Fix/clear trash on tras uses case

JoanVicens:
#### Trash Module
- Moved the `clearTrash` logic to trash use cases.
- `clear trash` no longer waits for the operation to finish
#### Added the following use cases needed to delete files and folders:
- Added remove orphans folders use case
- Added remove file shares use case

### DIFF
--- a/src/modules/trash/trash.controller.ts
+++ b/src/modules/trash/trash.controller.ts
@@ -97,13 +97,11 @@ export class TrashController {
   }
 
   @Delete('/all')
-  @HttpCode(204)
+  @HttpCode(202)
   @ApiOperation({
     summary: "Deletes all items from user's trash",
   })
-  async clearTrash(@UserDecorator() user: User) {
-    await this.trashUseCases.clearTrash(user);
-
-    return;
+  clearTrash(@UserDecorator() user: User) {
+    this.trashUseCases.clearTrash(user);
   }
 }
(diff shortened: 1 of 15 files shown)


src/modules/trash/trash.controller.ts
```diff
@@ -97,27 +97,11 @@ export class TrashController {
...
-      await this.folderUseCases
-        .getChildrenFoldersToUser(folderId, userId, deleted)
-        .then((folders: Array<Folder>) =>
-          folders.map(this.folderUseCases.deleteFolderPermanently),
-        ),
-    ]);
-
-    return;
+  clearTrash(@UserDecorator() user: User) {
+    this.trashUseCases.clearTrash(user);
```
fabioespinosa, line 105: perhaps you need to use `return` here?

src/modules/trash/trash.usecase.ts
```diff
@@ -0,0 +1,55 @@
...
+
+@Injectable()
+export class TrashUseCases {
+  constructor(
+    private fileUseCases: FileUseCases,
+    private folderUseCases: FolderUseCases,
+  ) {}
+
+  private async deleteFiles(files: Array<File>, user: User) {
+    await Promise.all(
```
fabioespinosa, line 16: try to avoid using Promise.all because if there are many files to delete, it may crash the event loop.

Better to delete them sequentially using for/await or to use async.queue

src/modules/trash/trash.controller.ts
JoanVicens: I did search it and I couldn't find any reason to have to return void

src/modules/trash/trash.usecase.ts
JoanVicens: Updated

JoanVicens:
⚠️ For this release the `STORAGE_API_URL` environment variable will be needed

PR #51, opened 2022-08-29
[WT-29] Feature: delete items from trash

JoanVicens:
Added delete multiple items from trash endpoint.

Depends on #50


JoanVicens:
I'm closing this because i was unable to run the unit test in this branch, #60 has the same changes

PR #61, opened 2022-08-31
[_]Fix: ssl connection on test

JoanVicens:
ssl configuration for development and test.

### DIFF
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -53,12 +53,14 @@ import { BridgeModule } from './externals/bridge/bridge.module';
           idle: 20000,
           acquire: 20000,
         },
-        dialectOptions: {
-          ssl: {
-            require: true,
-            rejectUnauthorized: false,
-          },
-        },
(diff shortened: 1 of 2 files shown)


src/app.module.ts
```diff
@@ -53,12 +53,14 @@ import { BridgeModule } from './externals/bridge/bridge.module';
           idle: 20000,
           acquire: 20000,
         },
-        dialectOptions: {
-          ssl: {
-            require: true,
-            rejectUnauthorized: false,
-          },
-        },
+        dialectOptions: configService.get('isProduction')
```
fabioespinosa, line 56: I think this gets this out of the way, but we should make this file with as little logic as possible. perhaps do something like configService.get('dialectOptions') and have the logic of assigning this in configuration.ts

Or perhaps we can have this as an environment variable like: RDS_REQUIRE_SSL and set that to ssl -> require

Otherwise we might end with a bunch of logic in app.module and is not easy to see where we mean production/development, etc...

PR #71, opened 2022-09-15
Develop

foxkdev:


fabioespinosa (approved):
I approve but I understand the endpoint is still very slow

PR #75, opened 2022-09-19
[WT-76] feat: add password field to shared links

JoanVicens:
Adds a nullable password field to shares and send links table

### DIFF
--- a/migrations/20220919083117-add-password-to-shared-links.js
+++ b/migrations/20220919083117-add-password-to-shared-links.js
@@ -2,20 +2,20 @@
 
 module.exports = {
   async up(queryInterface, Sequelize) {
-    await queryInterface.addColumn('shares', 'password', {
+    await queryInterface.addColumn('shares', 'hashed_password', {
       type: Sequelize.BLOB('medium'),
       allowNull: true,
     });
 
-    await queryInterface.addColumn('send_links', 'password', {
+    await queryInterface.addColumn('send_links', 'hashed_password', {
       type: Sequelize.BLOB('medium'),
       allowNull: true,
     });
   },
 
   async down(queryInterface) {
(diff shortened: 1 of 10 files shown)


fabioespinosa (approved):
Nicely done

migrations/20220919083117-add-password-to-shared-links.js
```diff
@@ -2,20 +2,20 @@
 
 module.exports = {
   async up(queryInterface, Sequelize) {
-    await queryInterface.addColumn('shares', 'password', {
+    await queryInterface.addColumn('shares', 'hashed_password', {
       type: Sequelize.BLOB('medium'),
```
fabioespinosa, line 6: I think we can use 'text' and save it in base64. Any reason to use Blob?

JoanVicens: any, used blob to match with the password field on user, bus since  is the hashed there is no need indeed.

src/modules/send/send-link-item.domain.ts
```diff
@@ -71,7 +71,6 @@ export class SendLinkItem implements SendLinkItemAttributes {
       createdAt: this.createdAt,
       updatedAt: this.updatedAt,
       version: this.version,
-      hashedPassword: this.hashedPassword,
```
JoanVicens: removed the `hashedPassword` from here since there are endpoints that return the serialized model

src/modules/share/share.domain.ts
```diff
@@ -134,7 +134,6 @@ export class Share implements ShareAttributes {
       active: this.active,
       createdAt: this.createdAt,
       updatedAt: this.updatedAt,
-      hashedPassword: this.hashedPassword,
```
JoanVicens: same as before

PR #77, opened 2022-09-21
[WT-95] fix/initialize crypto service

JoanVicens:
When the crypto service was created before the config service loaded the .env file the process environment where undefined making the crypto sevice crash:

After explore a few options decided to not decrypt the name of files and folders until is necessary. This changes de value of name on the domain layer to be an encrypted string that will need to be decrypted before sending the file or folder info to the client application.

Problems found with the others options:

- Export app:
When running unit test, the application starts every time file or folder domain is loaded.

- Domain builder (create a builder with crypto service injected that decrypts the name):
   1. Only models can be injected to Sequelize repositories
   2. Since the is an external service in charge of file and folder creation there is no easy solution to ensure that the domain object is created correctly with the constructor or a builder method.

Also removed the instantiation of the services as a singleton since <s>nestjs already does that for services:</s> https://docs.nestjs.com/providers#dependency-injection (thxs to @fabioespinosa for pointing that out)

Also: the test folder > calls getChildrenFoldersToUser and return folders is failing, but can't see why the result should be null event without the changes to Folder (domain)


src/modules/folder/folder.usecase.ts
```diff
@@ -127,4 +129,13 @@ export class FolderUseCases {
       await this.deleteOrphansFolders(userId);
     }
   }
+
+  decryptFolderName(folder: Folder): any {
```
fabioespinosa, line 133: Do we have this as well for files?

As in file.usecase.ts should have a decryptName

JoanVicens: Yes,should have it too. Adding it

fabioespinosa:
DI does involve singletons by default, but its 1 singleton per module, unless you pass around the instance via exports

Worth a read: https://blog.devgenius.io/nestjs-use-a-solo-injectable-class-as-a-singleton-provider-884911eff279

JoanVicens:
Another thing to consider is the creation of files and folders, currently there is no logic for the creation but if we added it the process would be something like:
```
const file = File.build()
fileRepository.create(file)
```
So we would be storing the file with the name decrypted and when its retrieved would be decrypted again

src/modules/folder/folder.usecase.ts
JoanVicens: Added!

PR #85, opened 2022-09-25
[_]: feat/signup

sg-gs:
Migrates the functionalities required for doing signup ("creating a user"), this includes:
- Analytics and tracks are sent when a user is created.
- Calls to the storage API to create a user there.
- Events emitted to the Notifications service. 
- Creating the referrals of the new user. 
- Cryptographic methods required to perform the signup
- Auth middlewares
- The payments service to be able to ask our providers if some user has already a subscription.
- The referral application when he/she signs up with a given referral code.

### DIFF
--- a/src/externals/bridge/bridge.service.ts
+++ b/src/externals/bridge/bridge.service.ts
@@ -1,9 +1,9 @@
 import { Logger } from '@nestjs/common';
 import { Inject, Injectable } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
-import { rejectedSyncPromise } from '@sentry/utils';
+
 import { FileAttributes } from 'src/modules/file/file.domain';
-import { User } from 'src/modules/user/user.domain';
+import { User, UserAttributes } from 'src/modules/user/user.domain';
 import { CryptoService } from '../crypto/crypto';
 import { HttpClient } from '../http/http.service';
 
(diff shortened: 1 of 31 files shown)


src/externals/bridge/bridge.service.ts
```diff
@@ -48,4 +103,23 @@ export class BridgeService {
       params,
     );
   }
+
+  async addStorage(
+    email: UserAttributes['bridgeUser'],
```
fabioespinosa, line 108: We also want to provide an `addStorage` by UUID here, referrals is using it in old drive-web

src/modules/user/dto/create-user.dto.ts
```diff
@@ -0,0 +1,76 @@
...
+    description: 'Salt',
+  })
+  salt: string;
+
+  @IsOptional()
+  @ApiProperty({
+    example: '',
+    description: '',
+  })
+  privateKey: string;
```
fabioespinosa, line 54: Why is the privateKey optional?

src/modules/user/user.usecase.ts
```diff
@@ -27,4 +68,217 @@ export class UserUseCases {
...
+    const userReferralsToCreate: Omit<
+      UserReferralAttributes,
+      'id' | 'startDate'
+    >[] = [];
+
+    referrals.forEach((referral) => {
+      Array(referral.steps)
+        .fill(null)
+        .forEach(() => {
+          const applied = referral.key === 'create-account';
```
fabioespinosa, line 272: We should centralize the hardcoded values at least in an enum 'create-account', 'complete-survey' and so on

src/externals/bridge/bridge.service.ts
sg-gs: Good point

src/modules/user/dto/create-user.dto.ts
sg-gs: Currently is not being sent to all the clients. In the login endpoint, the keys are generated in case they do not exist already.

src/externals/bridge/bridge.service.ts
sg-gs: Fixed

src/modules/user/user.usecase.ts
sg-gs: Done

fabioespinosa (approved):

PR #91, opened 2022-10-06
[WT-100]: Create migration for Thumbnails

larry-internxt:
- Added migration for Thumbnails

### DIFF
--- a/migrations/20221003090000-create-thumbnails-table.js
+++ b/migrations/20221003090000-create-thumbnails-table.js
@@ -1,0 +1,72 @@
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.createTable(
+      'thumbnails',
+      {
+        id: {
+          type: Sequelize.INTEGER,
+          primaryKey: true,
+          autoIncrement: true,
+          allowNull: false,
+        },
+        file_id: {
+          type: Sequelize.INTEGER,
+          allowNull: false,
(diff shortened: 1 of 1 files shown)


migrations/20221003090000-create-thumbnails-table.js
```diff
@@ -0,0 +1,72 @@
...
+        },
+        updated_at: {
+          type: Sequelize.DATE,
+          defaultValue: Sequelize.fn('now'),
+        },
+      },
+      {
+        uniqueKeys: {
+          unique_thumbnail: {
+            fields: ['file_id', 'type'],
```
fabioespinosa, line 46: This would cause conflict if we wanted more than one thumbnail per file right?

migrations/20221003090000-create-thumbnails-table.js
```diff
@@ -0,0 +1,72 @@
...
+        field: 'id',
+      },
+      onDelete: 'CASCADE',
+      onUpdate: 'CASCADE',
+    });
+  },
+
+  down: async (queryInterface) => {
+    await queryInterface.dropTable('thumbnails');
+    await queryInterface.removeConstraint(
```
fabioespinosa, line 67: Remove the constraint before dropping the table first, otherwise I believe it will throw an error to remove a constraint of a non existing table

fabioespinosa (approved):

PR #110, opened 2022-11-04
[_]: Fix/Add trash filter to shared items

larry-internxt:
- Added trash filter to shared items

### DIFF
--- a/src/modules/file/file.repository.ts
+++ b/src/modules/file/file.repository.ts
@@ -82,6 +82,12 @@ export class FileModel extends Model implements FileAttributes {
   updatedAt: Date;
 }
 
+interface FileParams {
+  fileId: FileAttributes['id'];
+  userId: FileAttributes['userId'];
+  deleted: FileAttributes['deleted'];
+}
+
 export interface FileRepository {
   findAll(): Promise<Array<File> | []>;
   findAllByFolderIdAndUserId(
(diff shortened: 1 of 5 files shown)


src/modules/share/share.repository.ts
```diff
@@ -142,7 +158,15 @@ export class SequelizeShareRepository implements ShareRepository {
...
     const share = await this.shareModel.findOne({
       where: { fileId, userId },
-      include: [this.fileModel, this.userModel],
+      include: [
+        this.userModel,
+        {
+          model: this.fileModel,
+          where: {
+            deleted: false,
+          },
```
fabioespinosa, line 167: Why do the others have required: false and this one doesn't ?

What does required do?

src/modules/file/file.repository.ts
```diff
@@ -94,6 +94,7 @@ export interface FileRepository {
   findOne(
     fileId: FileAttributes['id'],
     userId: FileAttributes['userId'],
+    deleted: FileAttributes['deleted'],
```
fabioespinosa, line 97: having a boolean in the parameters like this is usually not so readable when we use the function if you want to pass a filter make it an object so that when  call findOne it can be:

`const file = await this.filesRepository.findOne(fileId, user.id, {deleted: false});`

instead of:
`const file = await this.filesRepository.findOne(fileId, user.id, false);`

If someone is reading the latter, they have no idea what false is for and they have to check the function/types

fabioespinosa (changes requested):

fabioespinosa (approved):

PR #111, opened 2022-11-08
[_]: Fix/delete permanently trash items

larry-internxt:
- Fixed deleting permanently trash items bug

### DIFF
--- a/src/modules/folder/folder.usecase.ts
+++ b/src/modules/folder/folder.usecase.ts
@@ -26,8 +26,11 @@ export class FolderUseCases {
     private readonly cryptoService: CryptoService,
   ) {}
 
-  async getFolder(folderId: FolderAttributes['id']) {
-    const folder = await this.folderRepository.findById(folderId);
+  async getFolder(
+    folderId: FolderAttributes['id'],
+    deleted: FolderAttributes['deleted'] = false,
+  ) {
+    const folder = await this.folderRepository.findById(folderId, deleted);
     if (!folder) {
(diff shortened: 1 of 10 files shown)


src/modules/folder/folder.usecase.ts
```diff
@@ -26,8 +26,11 @@ export class FolderUseCases {
     private readonly cryptoService: CryptoService,
   ) {}
 
-  async getFolder(folderId: FolderAttributes['id']) {
-    const folder = await this.folderRepository.findById(folderId);
+  async getFolder(
+    folderId: FolderAttributes['id'],
+    deleted: FolderAttributes['deleted'] = false,
```
fabioespinosa, line 31: Having a boolean in a parameter without naming is a code smell, if you want to pass options to this method it is better advised to have it:

async getFolder(folderId, options);

Where options can have a {deleted: boolean} parameter.

Otherwise you can have someone use this method as: getFoler(<an id>, false); and someone new to the code will have to look inside to understand what is going on.

Better to have: 

getFolder(<an id>, {deleted: false}) so it is very clear what is going on.

Even though below you are calling it with an extra variable `deleted` , this behavior is not enforced and someone else might use it differently. 

I know having just one parameter is not such a big deal, but ultimately this can lead to having a method like:

getFolder(<an id>, false, true, false, false, true);

And then we have no idea what is going on, plus it's a source of bugs with the order

fabioespinosa (changes requested):
function signature

fabioespinosa (approved):
All good from my side

PR #119, opened 2022-11-22
[WT-182]: bugfix/Add password length limit

larry-internxt:
- Added password length limit

### DIFF
--- a/src/modules/user/user.usecase.ts
+++ b/src/modules/user/user.usecase.ts
@@ -5,6 +5,7 @@ import { v4 } from 'uuid';
 
 import { SequelizeUserRepository } from './user.repository';
 import {
+  MAX_PASSWORD_LENGTH,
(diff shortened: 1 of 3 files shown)


src/modules/user/user.usecase.ts
```diff
@@ -183,6 +192,10 @@ export class UserUseCases {
     const userPass = this.cryptoService.decryptText(password);
     const userSalt = this.cryptoService.decryptText(salt);
 
+    if (userPass.length > MAX_PASSWORD_LENGTH) {
+      throw new PasswordExceededLimitLengthError();
```
fabioespinosa, line 196: I believe you are validating the hashed password not the submitted 🤔 

Correct me if i'm wrong though

fabioespinosa (changes requested):

PR #120, opened 2022-11-25
[WT-216] Fix: server key not stored on user creation

JoanVicens:
Server Keys were not stored on user creation. Even they are present on the request DTO


src/modules/keyserver/key-server.usecase.ts
```diff
@@ -0,0 +1,35 @@
...
+
+@Injectable()
+export class KeyServerUseCases {
+  constructor(
+    @Inject('KEY_SERVER_REPOSITORY')
+    private readonly repository: KeyServerRepository,
+  ) {}
+
+  async addKeysToUser(userId: UserAttributes['id'], keys: Keys): Promise<Keys> {
+    if (!keys.publicKey) {
```
sg-gs, line 14: These checks should be on the controller, not in the use case, as Keys are not an optional on the interface too, therefore the checks are not correct according to the types, which indicates that those params are not null or undefined

sg-gs (changes requested):

src/modules/keyserver/key-server.usecase.ts
JoanVicens: Removed

sg-gs (approved):

JoanVicens:
@PixoDev commented that on mobile we don't send the keys. So we need to check if they are present or not in order to save them

sg-gs (approved):

PR #122, opened 2022-12-02
[_]: Fix/Add trycatch to ignore errors on signup

larry-internxt:
- Added trycatch to ignore errors on signup


fabioespinosa:
As referrals and sending email is not critical to signing up, lgtm

fabioespinosa (approved):

PR #135, opened 2023-02-02
[_]: fix/allow-failed-signups-to-register

sg-gs:
These changes allow a failed signup to signup again. 

**_Issue_**
We create one account on Drive and one in our network for each signup. When the network account was created successfully but the Drive one did not, the user was not able to signup again as the network was rejecting the email as it already existed. 

**_Solution_**
Calling to a gateway endpoint on the network that creates or updates the user if already exists. Therefore, if the signup fails, as long as the Drive account does not exist, the user will be able to signup again.


sg-gs:
Note: The environment variable `GATEWAY_SECRET` is already set on production

PR #148, opened 2023-02-27
[WT-340]: Feat/Add lastPasswordChangeAt to users

larry-internxt:
- Added `lastPasswordChangeAt `column to users
- Added middleware to unauthorize users when `lastPasswordChangedAt` is more recent than token `iat`.
- Added unit tests to jwt strategy.


migrations/20230227160000-add-lastPasswordChangedAt-to-users.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.addColumn('users', 'last_password_changed_at', {
+      type: Sequelize.DATE,
+      allowNull: true,
```
sg-gs, line 8: same comment as on the drive-server PR regarding null values

apsantiso:
@sg-gs I checked this PR is also related with PB-267 which is still in progress, can we accept this PR as it is and add the middleware later? 

We could have a simple function to check the token that we can use both in the middleware and in the endpoint for https://github.com/internxt/drive-server-wip/pull/250.  

Public endpoints (as auto unblock account) don´t pass through the middleware we don´t rely on the implementation that is still in progress.

what do you think?

migrations/20240109071854-add-lastPasswordChangedAt-to-users.js
```diff
@@ -0,0 +1,23 @@
...
+
+const tableName = 'users';
+const newColumn = 'last_password_changed_at';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    const tableDefinition = await queryInterface.describeTable(tableName);
+    //  Only add column if is not created in the table
+    if (!tableDefinition[newColumn]) {
```
sg-gs, line 11: The migrations should be idempotent, not dependent on the environment. Do not worry about production, we will run a clean migration.

sg-gs, line 11: The migrations should be idempotent, not dependent on the environment. Do not worry about production, we will run a clean migration.

src/modules/auth/jwt.strategy.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    jest.spyOn(userUseCases, 'getUser').mockResolvedValue(null);
+
+    await expect(
+      strategy.validate({ payload: { uuid: 'anyUuid' } }),
+    ).rejects.toThrow(UnauthorizedException);
+  });
+
+  it('When token iat is older than lastPasswordChangedAt , then fail', async () => {
+    const user = newUser();
+    const greaterDate = new Date();
```
sg-gs, line 36: To ensure that this case is not failing, please, do use the token iat as the base date for a new Date, and the add the minutes. If the tests start to run slowly, this test could not pass. The tests should be consistent always. For instance: 

```
const user = newUser();
const tokenIat = getTokenDefaultIat();
const olderDate = new Date(tokenIat);
olderDate.setMinutes(olderDate.getMinutes() + 1);
// and so on ..
```

src/modules/auth/jwt.strategy.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    await expect(
+      strategy.validate({ payload: { uuid: 'anyUuid' }, iat: tokenIat }),
+    ).rejects.toThrow(UnauthorizedException);
+  });
+
+  it('When user has lastPasswordChangedAt older than token iat, then return user', async () => {
+    const tokenIat = getTokenDefaultIat();
+    const user = newUser();
+    const olderDate = new Date();
```
sg-gs, line 51: Same here. Switching the orders of both things could make the tests pass. The control of the things that determine if the tests passes or not should be clear here. Ideally, we should not have to check the internal code to be sure that the tests works as expected

sg-gs (changes requested):

sg-gs (approved):

PR #149, opened 2023-02-28
[_]: refactor/delete-items-permanently

sg-gs:


src/externals/bridge/bridge.service.ts
```diff
@@ -113,10 +113,14 @@ export class BridgeService {
...
     );
 
-    await this.httpClient.delete(
-      `${url}/buckets/${bucket}/files/${bucketEntry}`,
-      params,
-    );
+    await this.httpClient
+      .delete(`${url}/buckets/${bucket}/files/${bucketEntry}`, params)
+      .catch((err) => {
+        if (err.response.status === 404) {
```
sg-gs, line 119: This is the fix for emptying the trash that was failing @CandelR

src/modules/trash/trash.controller.ts
```diff
@@ -206,8 +206,23 @@ export class TrashController {
   @ApiOperation({
     summary: "Deletes all items from user's trash",
   })
-  clearTrash(@UserDecorator() user: User) {
-    this.trashUseCases.clearTrash(user);
```
sg-gs, line 210: Trash issue where the trash was not emptied but the response from the server is given, which causes 'ghost' trash items to appear and disappear @CandelR @PixoDev just FYI

PR #163, opened 2023-06-09
[PB-535]: feat/return list of folders shared with a user

cg27shokworks:
Return list of folders shared with a user
Include folder and shares seeds


src/modules/share/share.controller.ts
```diff
@@ -238,6 +245,55 @@ export class ShareController {
...
+  @ApiOkResponse({ description: 'Get all folders shared with a user' })
+  @ApiBearerAuth()
+  async getSharedFolders(
+    @UserDecorator() user: User,
+    @Query('page') page: number,
+    @Query('perPage') perPage: number,
+    @Query('orderBy') orderBy: OrderBy,
+  ) {
+    try {
+      const Folders = await this.shareUseCases.getSharedFoldersToAUser(
```
sg-gs, line 280: use camelCase please

src/modules/share/share.repository.ts
```diff
@@ -238,6 +243,57 @@ export class SequelizeShareRepository implements ShareRepository {
...
+    perPage: number,
+    orderBy?: OrderBy,
+  ): Promise<Folder[]> {
+    const { offset, limit } = Pagination.calculatePagination(page, perPage);
+
+    const validTypePaginationParameters =
+      (isNaN(page) && !isNaN(perPage)) || (!isNaN(page) && isNaN(perPage));
+
+    if (validTypePaginationParameters) {
+      throw new BadRequestException('take and skip must be provided together');
```
sg-gs, line 258: Let's handle HTTP things on the controller layer please

src/modules/share/share.repository.ts
```diff
@@ -238,6 +243,57 @@ export class SequelizeShareRepository implements ShareRepository {
...
+    }
+
+    const order = orderBy
+      ? [orderBy.split(':') as [string, string]]
+      : undefined;
+
+    const findCriteria = {
+      where: {
+        user_id: user.id,
```
sg-gs, line 267: use camelCase

src/modules/share/share.repository.ts
```diff
@@ -238,6 +243,57 @@ export class SequelizeShareRepository implements ShareRepository {
...
+
+    if (validTypePaginationParameters) {
+      throw new BadRequestException('take and skip must be provided together');
+    }
+
+    const order = orderBy
+      ? [orderBy.split(':') as [string, string]]
+      : undefined;
+
+    const findCriteria = {
```
sg-gs, line 265: I will type this if we are going to split it from the `findAll`

src/modules/share/share.repository.ts
```diff
@@ -238,6 +243,57 @@ export class SequelizeShareRepository implements ShareRepository {
...
+          model: this.folderModel,
+          where: {
+            deleted: false,
+          },
+          required: false,
+        },
+      ],
+    };
+
+    if (!isNaN(limit) && !isNaN(offset)) {
```
sg-gs, line 281: The repository layer should persist data only, the responsibility of validating the data should be done by the controller. All the controllers must validate the data so any usecase/repository receives the data in the expected format. If we have types for variables but we still have to validate things on the entire codebase, then TypeScript is not worth it

src/modules/share/share.repository.ts
```diff
@@ -238,6 +243,57 @@ export class SequelizeShareRepository implements ShareRepository {
...
+      findCriteria['offset'] = offset;
+    }
+
+    if (order) {
+      findCriteria['order'] = order;
+    }
+
+    const shares = await this.shareModel.findAll(findCriteria);
+
+    const folders = shares.map((share) => share.folder.toJSON());
```
sg-gs, line 292: return directly

src/modules/share/share.repository.ts
```diff
@@ -238,6 +243,57 @@ export class SequelizeShareRepository implements ShareRepository {
...
+    if (!isNaN(limit) && !isNaN(offset)) {
+      findCriteria['limit'] = limit;
+      findCriteria['offset'] = offset;
+    }
+
+    if (order) {
+      findCriteria['order'] = order;
+    }
+
+    const shares = await this.shareModel.findAll(findCriteria);
```
sg-gs, line 290: According to the confluence documentation, we should return **the private folders** shared by a user with another user, shares table is for public sharing

sg-gs (changes requested):
I recommend re-checking the Confluence documentation regarding this task first. We have a different domain for private sharing but this codebase is working on the public sharing domain

cg27shokworks:
I will close this PR to open another one based on https://github.com/internxt/drive-server-wip/pull/168 changes.

PR #165, opened 2023-06-14
[PB-572 ]: bugfix/Preview thumbnails are not rendered in gallery view

CandelR:
- Added thumbnails module
- Added shares and thumbnails to files
- Separated file model from respository


src/modules/file/file.repository.ts
```diff
@@ -205,6 +109,24 @@ export class SequelizeFileRepository implements FileRepository {
       limit,
       offset,
       where,
+      include: [
```
sg-gs, line 112: This should be optional, we do not want to do queries that make things slower without any reason

src/modules/thumbnail/thumbnail.attributes.ts
```diff
@@ -0,0 +1,13 @@
+export interface ThumbnailAttributes {
+  id: number;
+  fileId: number;
+  type: string;
+  size: number;
+  bucket_id: string;
```
sg-gs, line 6: We should use Sequelize uppercase feature to make this camelCase automatically

src/modules/thumbnail/thumbnail.model.ts
```diff
@@ -0,0 +1,57 @@
...
+export class ThumbnailModel extends Model implements ThumbnailAttributes {
+  @PrimaryKey
+  @Column
+  id: number;
+
+  @ForeignKey(() => FileModel)
+  @Column(DataType.INTEGER)
+  fileId: number;
+
+  @BelongsTo(() => FileModel, 'fileId')
```
sg-gs, line 28: We relate `thumbnails.fileId` with `files.id`
`@BelongsTo(() => FileModel, 'id')`

sg-gs (changes requested):

src/modules/file/file.repository.ts
CandelR: With the parameter 'required: false' below, isn't it optional?

src/modules/thumbnail/thumbnail.attributes.ts
CandelR: in the frontend it uses `bucket_id` and `bucket_file` in snake_case, I guess it was done this way because in drive-server it was returned this way. 
I imagine that in mobile it will be the same, I leave these two fields in snake_case to avoid unexpected errors?

src/modules/file/file.repository.ts
sg-gs: Required does [mean](https://sequelize.org/docs/v6/advanced-association-concepts/eager-loading/#required-eager-loading) that the query is an INNER or an OUTER JOIN, but the database, in both cases, is trying to get the results. The difference between them is that if is true, then the file should be returned only if the included models have a row related to the file (INNER) otherwise, the file will be always returned (OUTER). 

We are looking to be able to do a `required: false` of thumbnails if and only if we need it. Therefore, I would recommend adding a new method called `findAllCursorWithThumbnails` which does the job, a controller query param to request the inclusion, and an if on the use case that calls findAllCursor or findAllCursorWithThumbnails depending on that.

src/modules/thumbnail/thumbnail.attributes.ts
sg-gs: Leave them in both ways (snake, camel), but we should correct this on the frontend, the API should not depend on any platform implementation detail

src/modules/thumbnail/thumbnail.repository.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  async create(thumbnail: Thumbnail): Promise<Thumbnail> {
+    const thumbnailModel = this.toModel(thumbnail);
+    const createdThumbnail = await this.thumbnailModel.create(thumbnailModel);
+    return this.toDomain(createdThumbnail);
+  }
+
+  async update(thumbnail: Thumbnail): Promise<void> {
+    const thumbnailModel = await this.thumbnailModel.findByPk(thumbnail.id);
```
sg-gs, line 47: Use `updateOne`, which does one thing in one query, instead of in two

sg-gs (changes requested):

src/modules/thumbnail/thumbnail.attributes.ts
CandelR: Yes, I agree. It's just not to break anything that is already done by returning the fields in a different format, as the changes are required on the front end. Later on we can add a task to make the modification on the different platforms.

src/modules/file/file.usecase.ts
```diff
@@ -232,6 +251,21 @@ export class FileUseCases {
     }).toJSON();
   }
 
+  addOldAttributes(file: File): any {
+    const thumbnails = file.thumbnails;
+
+    const thumbnailsWithOldAttributers = thumbnails.map((thumbnail) => ({
+      ...thumbnail,
+      bucket_id: thumbnail.bucketId,
```
sg-gs, line 259: what is this?

sg-gs (approved):

PR #170, opened 2023-06-20
[PB-537]: feat/create private folder permission

cg27shokworks:
Adds endpoint to create private folder permission

### DIFF
--- a/src/modules/folder/folder.repository.ts
+++ b/src/modules/folder/folder.repository.ts
@@ -321,6 +321,17 @@ export class SequelizeFolderRepository implements FolderRepository {
     );
   }
 
+  async isOwner(user: User, folderId: number): Promise<boolean> {
+    const folder = await this.folderModel.findOne({
+      where: {
+        id: folderId,
+        userId: user.id,
+      },
+    });
+
+    return !!folder;
+  }
+
   async findAllCursorWhereUpdatedAfter(
     where: Partial<Folder>,
     updatedAfter: Date,
--- a/src/modules/private-share-folder/dto/grant-privilages.dto.ts
+++ b/src/modules/private-share-folder/dto/grant-privilages.dto.ts
@@ -1,0 +1,16 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { IsUUID } from 'class-validator';
+
+export class GrantPrivilegesDto {
+  @IsUUID()
+  @ApiProperty({ required: true })
+  userUuid: string;
+
+  @IsUUID()
+  @ApiProperty({ required: true })
+  privateFolderId: string;
+
+  @IsUUID()
+  @ApiProperty({ required: true })
+  roleId: string;
+}
--- a/src/modules/private-share-folder/private-sharing.controller.ts
+++ b/src/modules/private-share-folder/private-sharing.controller.ts
@@ -1,4 +1,11 @@
-import { Controller, Get, Query } from '@nestjs/common';
+import {
+  BadRequestException,
+  Body,
+  Controller,
+  Get,
+  Post,
+  Query,
+} from '@nestjs/common';
 import {
   ApiBearerAuth,
   ApiOkResponse,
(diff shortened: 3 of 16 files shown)


sg-gs:
Let's merge with master and refresh everything @cg27shokworks

src/modules/folder/folder.repository.ts
```diff
@@ -321,6 +321,17 @@ export class SequelizeFolderRepository implements FolderRepository {
     );
   }
 
+  async isOwner(user: User, folderId: number): Promise<boolean> {
```
sg-gs, line 324: This is a find by id which apart from existing, contains the name of something that seems to be business (therefore, responsibility of the use case) related

src/modules/private-share-folder/dto/grant-privilages.dto.ts
sg-gs, line 1: The file has a typo on its name, it should be grant-privileges

src/modules/private-share-folder/private-sharing.controller.ts
```diff
@@ -12,12 +19,39 @@ import { Folder } from '../folder/folder.domain';
...
+        dto.userUuid,
+        dto.privateFolderId,
+        dto.roleId,
+      );
+
+      return {
+        message: 'Privileges granted',
+      };
+    } catch (error) {
+      throw new BadRequestException(error.message);
```
sg-gs, line 51: Do not assume that this is always a bad exception error, also do not assume that the error message should be always exposed, by obvious security reasons

src/modules/private-share-folder/private-sharing.repository.ts
```diff
@@ -29,7 +32,37 @@ export class SequelizePrivateSharingRepository
...
+    private privateSharingFolderRole: typeof PrivateSharingFolderRolesModel,
   ) {}
+
+  async findById(id: string): Promise<PrivateSharingFolder> {
+    const privateFolder = await this.privateSharingFolderModel.findByPk(id);
+
+    return privateFolder.get({ plain: true });
+  }
+
+  async createPrivateFolder(
```
sg-gs, line 45: `createPrivateSharing` or just `create` could be a better naming

src/modules/private-share-folder/private-sharing.usecase.ts
```diff
@@ -2,12 +2,42 @@ import { Injectable } from '@nestjs/common';
...
+  ) {
+    const user = await this.userRespository.findByUuid(userUuid);
+    const privateFolder = await this.privateSharingRespository.findById(
+      privateFolderId,
+    );
+    const folder = await this.folderRespository.findByUuid(
+      privateFolder.folderId,
+    );
+
+    const isOwner = await this.folderRespository.isOwner(owner, folder.id);
```
sg-gs, line 29: There are unnecessary queries here. The user cames already loaded by the middleware, which by the way, already does query the user against the database. 

Due to that, I recommend to make it still secure but faster and easier to just query the folder by user id or just better, to ensure that when we finally migrate to user uuids definitely: 

Query the folder by the uuid and if it is the owner, just check that `user.id === folder.userId`

That requires just 2 queries, easier to maintain, to update and to test.

sg-gs: You will still have the user and folder objects required to create the private folder role of the user inside that folder

sg-gs (changes requested):

src/modules/private-share-folder/private-sharing.usecase.ts
```diff
@@ -2,12 +2,44 @@ import { Injectable } from '@nestjs/common';
...
 export class PrivateSharingUseCase {
   constructor(
     private privateSharingRespository: SequelizePrivateSharingRepository,
+    private userRespository: SequelizeUserRepository,
+    private folderRespository: SequelizeFolderRepository,
   ) {}
+  async grantPrivileges(
+    owner: User,
+    userUuid: string,
+    privateFolderId: string,
```
sg-gs, line 23: Could be a good idea to reference this type by the entity directly so in case the type of the entity field changes, you do not have ti change it on the entire codebase one by one. So instead of:

`privateFolderId: string`

you have:

`privateFolderId: Folder['uuid'] `

or even better:

`PrivateSharingFolder['folderId']`

sg-gs (changes requested):

sg-gs (approved):

PR #171, opened 2023-06-20
[PB-539]: feat/update private folder permission

cg27shokworks:
Endpoint to update private folder permissions


src/modules/private-share-folder/private-sharing.usecase.ts
```diff
@@ -38,6 +38,33 @@ export class PrivateSharingUseCase {
...
     );
   }
+
+  async updateRole(owner: User, privateFolderRoleId: string, roleId: string) {
+    const privateFolderRole =
+      await this.privateSharingRespository.findPrivateFolderRoleById(
+        privateFolderRoleId,
+      );
+
+    const folder = await this.folderRespository.findByUuid(
```
sg-gs, line 48: Same as here: https://github.com/internxt/drive-server-wip/pull/170#discussion_r1250904534

Also, may be worth checking that the role in fact exists, before trying to update it, just in case something is wrong, we notify the user and we also are able to log the error as expected

sg-gs (changes requested):

src/modules/private-share-folder/private-sharing.controller.ts
```diff
@@ -72,6 +77,50 @@ export class PrivateSharingController {
     }
   }
 
+  @Put('update-role/:id')
```
sg-gs, line 80: Could we change the naming of this? As a resource on REST, and as we use PUT:
`PUT /role/:id`
would be more correct

sg-gs (approved):

src/modules/private-share-folder/private-sharing.repository.ts
```diff
@@ -89,7 +89,7 @@ export class SequelizePrivateSharingRepository
     );
   }
 
-  async findPrivateFolderRoleByUserIdAndFolderId(
+  async findPrivateFolderRoleByFolderUuidAndUserUuid(
```
sg-gs, line 92: Hey @cg27shokworks, I will use userId and `folderId` as you are finding by, in this table, where the fields are called like that. It is precisely due to this and for maintenance reasons that using that naming would be better than this. I find the typing on TypeScript descriptive enough to reveal that currently, the value of those fields are the `user.uuid` and the `folder.uuid`

src/modules/private-share-folder/private-sharring.usecase.spec.ts
```diff
@@ -172,8 +167,15 @@ describe('PrivateSharingUseCase', () => {
...
+  });
+
+  describe('Manage private shared folder roles', () => {
+    const userUuid = v4();
+    const privateFolderId = v4();
+    const roleUuid = v4();
+    const owner = user;
 
-    it('should grant privileges if the owner is indeed the owner of the folder', async () => {
+    it('Grant privileges by owner', async () => {
```
sg-gs, line 178: Hey, check this legend, as an example: 
_When privileges are granted by the owner, then the invited user gets role_

sg-gs (changes requested):

miguelsw:
lgtm only fixes sergio's comments

PR #173, opened 2023-06-27
[PB-292] feat: fuzzy search

JoanVicens:


src/modules/fuzzy-search/fuzzy-search.controller.ts
```diff
@@ -0,0 +1,33 @@
...
+  @Get('/:search')
+  async fuzzySearch(
+    @UserDecorator() user: User,
+    @Param('search') search: string,
+  ) {
+    return await this.usecases.fuzzySearch(user.uuid, search);
+  }
+
+  @Put('/')
+  async add(@UserDecorator() user: User, @Body() content: RequestBody) {
```
sg-gs, line 26: We are going to add the item to the look up table once the resource is created (file/folder creation)

src/modules/fuzzy-search/look-up.model.ts
```diff
@@ -0,0 +1,18 @@
...
+  underscored: true,
+  tableName: 'look_up',
+})
+export class LookUpModel extends Model implements LookUpAttributes {
+  @PrimaryKey
+  @Column
+  id: string;
+
+  @Column
+  userUuid: string;
```
sg-gs, line 14: Reference User['uuid'] field as the type instead, please. Also let's call the userUuid column as userId as we are going to deprecate the auto incremental id at some point

src/modules/fuzzy-search/look-up.model.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  @Column
+  name: string;
```
sg-gs, line 17: Reference `File['plainName'] | Folder['plainName']` as the type please

src/modules/fuzzy-search/look-up.repository.ts
```diff
@@ -0,0 +1,56 @@
...
+  ): Promise<Array<LookUpAttributes>> {
+    const query = `
+      SELECT
+        *,
+        rank_name,
+        similarity
+      FROM
+        look_up as lu,
+        to_tsvector(lu."name") document,
+        to_tsquery('${partialName}') query,
```
sg-gs, line 37: Escape all the strings using [Sequelize Replacements](https://sequelize.org/docs/v6/core-concepts/raw-queries/#replacements) please

migrations/20230627120414-create-look-up-table.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const fuzzySearchTableName = 'look_up';
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.createTable(fuzzySearchTableName, {
```
sg-gs, line 7: I will relate `file.uuid` to `look_up.item_id` here so we are able to keep the table clean with `ON_DELETE=CASCADE` from files / folders via FK

sg-gs (approved):

migrations/20230627120414-create-look-up-table.js
JoanVicens: but item_id cannot reference files.uuid and folder.uuid at the same time. Do I create a look_up.file_id and look_up.folder_id to have the references?

src/modules/fuzzy-search/look-up.model.ts
JoanVicens: Changed on the attributes interface. The model implements it so it's centralized

migrations/20230627120414-create-look-up-table.js
```diff
@@ -0,0 +1,51 @@
...
+      name: {
+        type: Sequelize.STRING,
+        allowNull: false,
+      },
+      tokenized_name: {
+        type: Sequelize.DataTypes.TSVECTOR,
+        allowNull: false,
+      },
+      item_id: {
+        type: Sequelize.STRING(36),
```
sg-gs, line 22: The user_id has to be `Sequelize.DataTypes.STRING` because the users' table was created like that, but in the case of the files/folders, the type is the proper one: `Sequelize.DataTypes.UUID`, could you adjust it?

src/modules/fuzzy-search/fuzzy-search.controller.ts
```diff
@@ -0,0 +1,30 @@
...
+export class FuzzySearchController {
+  constructor(private readonly usecases: FuzzySearchUseCases) {}
+
+  @Get('/:search')
+  @ApiOperation({
+    summary: 'Search for items from a part of the name',
+  })
+  @HttpCode(200)
+  @ApiOkResponse({
+    description: 'Elements founded',
```
sg-gs, line 19: typo here, is 'found', not founded

sg-gs (changes requested):

sg-gs:
Could you also add the query to refill the table for files/folders that do not exist there still, as a migration, please? @JoanVicens 

Also, could you add a trigger for when files/folders are updated by name, to reindex this?

PR #175, opened 2023-06-29
[PB-542]: feat/update private key re encrypted

cg27shokworks:


src/modules/keyserver/key-server.repository.ts
```diff
@@ -64,4 +64,16 @@ export class SequelizeKeyServerRepository implements KeyServerRepository {
       defaults: data,
     });
   }
+
+  async update(
+    userId: UserAttributes['id'],
+    data: Partial<KeyServerAttributes>,
+  ) {
+    const [keyServer] = await this.findUserKeysOrCreate(userId, data);
+    if (!keyServer) {
```
sg-gs, line 73: I would let the use case handle this, as it is the use case imposing that you need to have the keys before updating. In some cases, this could not be important and therefore, not managed by the use case, but in this case, it is vital to ensure and report to the user that the keys are missing.

src/modules/user/user.controller.ts
```diff
@@ -150,4 +156,18 @@ export class UserController {
...
+  @ApiBearerAuth()
+  async updatePassword(
+    @RequestDecorator() req,
+    @Body() updatePasswordDto: UpdatePasswordDto,
+  ) {
+    try {
+      await this.userUseCases.updatePassword(req.user, updatePasswordDto);
+      return { status: 'success' };
+    } catch (err) {
+      throw new BadRequestException(err.message);
```
sg-gs, line 170: I would check first which is the error rather than assuming that it is a bad exception always, internal server exceptions, as the more clear case, should not return the original message which may expose delicate details about something in terms of security.

src/modules/user/user.usecase.ts
```diff
@@ -518,4 +558,36 @@ export class UserUseCases {
...
+
+  async updatePassword(
+    user: User,
+    updatePasswordDto: UpdatePasswordDto,
+  ): Promise<void> {
+    const { currentPassword, newPassword, newSalt, mnemonic, privateKey } =
+      updatePasswordDto;
+
+    if (user.password.toString() !== currentPassword) {
+      throw new InvalidPasswordError();
```
sg-gs, line 570: An UnauthorizedException could be more descriptive

src/modules/user/user.usecase.ts
```diff
@@ -518,4 +558,36 @@ export class UserUseCases {
...
+    }
+    await this.userRepository.updateById(user.id, {
+      password: newPassword,
+      hKey: new Buffer(newSalt),
```
sg-gs, line 574: Is the new Buffer usage on drive-server update password functionality? `new Buffer` is currently [deprecated](https://nodejs.org/api/buffer.html#new-bufferarray). Use `Buffer.from(data, 'format')` instead

src/modules/user/user.usecase.ts
```diff
@@ -518,4 +558,36 @@ export class UserUseCases {
...
+        encryptVersion: privateKey,
+      },
+    );
+
+    if (!keyServer) {
+      throw new KeyServerNotFoundError();
+    }
+
+    await this.keyServerRepository.update(user.id, {
+      encryptVersion: privateKey,
```
sg-gs, line 590: Encrypt version indicates the algorithm used for encrypting the `privateKey`, the `privateKey` has its own field called privateKey.

sg-gs (changes requested):

src/modules/user/user.usecase.ts
```diff
@@ -518,4 +555,36 @@ export class UserUseCases {
       password,
     });
   }
+
+  async updatePassword(
```
sg-gs, line 559: The current endpoint for this on drive-server is [here](https://github.com/internxt/drive-server/blob/dca8b2b42d1510cd30f71efd6e67e0c504307856/src/app/routes/user.js#L11) you will see that `currentPassword`, `newPassword`, and `newSalt` are encrypted on transit, so you have to decrypt them before using them.

src/modules/user/user.usecase.ts
```diff
@@ -518,4 +555,36 @@ export class UserUseCases {
...
+    user: User,
+    updatePasswordDto: UpdatePasswordDto,
+  ): Promise<void> {
+    const { currentPassword, newPassword, newSalt, mnemonic, privateKey } =
+      updatePasswordDto;
+
+    if (user.password.toString() !== currentPassword) {
```
sg-gs, line 566: The `toString()` is part of ensuring the format and should be handled by the controller, which ensures the format is correct (in this case, coming from HTTP) before calling the use case.

src/modules/user/user.usecase.ts
```diff
@@ -518,4 +555,36 @@ export class UserUseCases {
...
+      throw new UnauthorizedException();
+    }
+    await this.userRepository.updateById(user.id, {
+      password: newPassword,
+      hKey: Buffer.from(newSalt),
```
sg-gs, line 571: Why? Check [this](https://github.com/internxt/drive-server/blob/dca8b2b42d1510cd30f71efd6e67e0c504307856/src/app/services/user.js#L342)

src/modules/user/user.usecase.ts
```diff
@@ -518,4 +555,36 @@ export class UserUseCases {
...
+    });
+
+    const [keyServer] = await this.keyServerRepository.findUserKeysOrCreate(
+      user.id,
+      {
+        encryptVersion: privateKey,
+      },
+    );
+
+    if (!keyServer) {
```
sg-gs, line 582: This is just impossible if you have just above a `findOrCreate`

src/modules/user/user.usecase.ts
```diff
@@ -518,4 +555,36 @@ export class UserUseCases {
...
+    await this.userRepository.updateById(user.id, {
+      password: newPassword,
+      hKey: Buffer.from(newSalt),
+      mnemonic,
+    });
+
+    const [keyServer] = await this.keyServerRepository.findUserKeysOrCreate(
+      user.id,
+      {
+        encryptVersion: privateKey,
```
sg-gs, line 578: The encryptVersion should be sent by the front, which chooses how to encrypt the key (zero-knowledge), also, this field should not have the value of the privateKey, there is a field for that, as I have already said [here](https://github.com/internxt/drive-server-wip/pull/175#discussion_r1269689781). 

I also do not advise you to create anything, just find it. The creation of the keys is just enforced on one point: the login. Having it on multiple points could be a troublesome

sg-gs (changes requested):

sg-gs (approved):

PR #185, opened 2023-07-24
[PB-529]: feat/ get public key from user email

cg27shokworks:


src/modules/user/user.usecase.ts
```diff
@@ -82,6 +82,10 @@ export class UserUseCases {
     return this.userRepository.findByUsername(email);
   }
 
+  getUserByEmailAddress(email: UserAttributes['email']) {
```
sg-gs, line 85: Use `findByUsername`, as multiple users could have the same email on sharedWorkspaces, the username is like the email, but it does not change when a user joins a shared workspace

src/modules/user/user.controller.ts
```diff
@@ -249,4 +251,26 @@ export class UserController {
...
+  @ApiBearerAuth()
+  @ApiOperation({
+    summary: 'Get public key by email',
+  })
+  @ApiParam({
+    name: 'email',
+    type: String,
+  })
+  async getPublicKeyByEmail(@Param('email') email: User['email']) {
+    const user = await this.userUseCases.getUserByEmailAddress(email);
```
sg-gs, line 266: Call `findByUsername` here

sg-gs (changes requested):

sg-gs (approved):

sg-gs:
We have conflicts with master @cg27shokworks

PR #203, opened 2023-08-24
[PB-904] feat/send-share-and-sharing-in-get-folders

miguelsw:
- frontend needs to know when a  folder was/is shared

Update:
- provide sharing details when it gets folders

Related to: https://github.com/internxt/drive-web/pull/789


src/modules/share/share.repository.ts
```diff
@@ -267,7 +263,7 @@ export class SequelizeShareRepository implements ShareRepository {
         },
       },
       include: [
-        this.userModel,
+        UserModel,
```
sg-gs, line 266: Why are we including in a generic pagination the user model always? This slows all the use cases relying on this

src/modules/share/share.repository.ts
```diff
@@ -276,7 +272,7 @@ export class SequelizeShareRepository implements ShareRepository {
           required: false,
         },
         {
-          model: this.folderModel,
+          model: FolderModel,
```
sg-gs, line 275: Same with this

miguelsw: It was like a solution for the circular dependency that it showed me when adding the relations between folder and share and private share since the injected model was only being used for relations, I will rollback this

apsantiso:
I think this is not needed anymore. I will close it, but please reopen it if you need.

apsantiso:
I think this is not needed anymore. I will close it, but please reopen it if you need.

PR #204, opened 2023-08-25
[PB-907] : fix/catch-errors-and-validation-in-array

miguelsw:
Issue: 
- it only showed status 200 and did not validate some values of the request body

Update:
- I add some validations and display the caught error in the response

Note:
- try that this change does not break what is used in drive-web


apsantiso:
We added a check in the DTO a while ago, the only thing we are missing is the fetch after updating the items status. 

If we need to check if all items where trashed in the future, then we can reopen this

PR #207, opened 2023-09-01
[PB-685]: fix/trash-not-being-emptied-properly

miguelsw:
Issue:
- Did not completely delete the files

update:
- Set to count folders to delete

### DIFF
--- a/src/modules/file/file.usecase.ts
+++ b/src/modules/file/file.usecase.ts
@@ -321,6 +321,8 @@ export class FileUseCases {
     return this.fileRepository.getFilesCountWhere({
       userId,
       status: FileStatus.TRASHED,
+      deleted: true,
+      removed: false,
     });
   }
 
--- a/src/modules/file/file.usecase.ts
+++ b/src/modules/file/file.usecase.ts
@@ -321,7 +321,6 @@ export class FileUseCases {
     return this.fileRepository.getFilesCountWhere({
       userId,
       status: FileStatus.TRASHED,
-      removed: false,
     });
   }
 
--- a/src/modules/file/file.usecase.ts
+++ b/src/modules/file/file.usecase.ts
@@ -321,7 +321,6 @@ export class FileUseCases {
     return this.fileRepository.getFilesCountWhere({
       userId,
       status: FileStatus.TRASHED,
-      deleted: true,
       removed: false,
     });
   }
--- a/src/modules/trash/trash.usecase.ts
+++ b/src/modules/trash/trash.usecase.ts
@@ -20,10 +20,14 @@ export class TrashUseCases {
    * @param trashOwner User whose trash is going to be emptied
    */
   async emptyTrash(trashOwner: User): Promise<void> {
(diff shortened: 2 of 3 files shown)


src/modules/trash/trash.usecase.ts
```diff
@@ -20,10 +20,14 @@ export class TrashUseCases {
    * @param trashOwner User whose trash is going to be emptied
    */
   async emptyTrash(trashOwner: User): Promise<void> {
-    const count = await this.fileUseCases.getTrashFilesCount(trashOwner.id);
+    const filesCount = await this.fileUseCases.getTrashFilesCount(
+      trashOwner.id,
+    );
+    const foldersCount = await this.folderUseCases.getTrashFoldersCount(
```
sg-gs, line 26: It seems like this method does not filter already removed folders. Please do fix it including `{ removed: false }` on the query of folderUseCases.[getTrashFoldersCount](https://github.com/internxt/drive-server-wip/blob/ca4febf2c0978f1617c5c4dc94b0fe050c9a475c/src/modules/folder/folder.usecase.ts#L450)

sg-gs (changes requested):

src/modules/file/file.usecase.ts
```diff
@@ -321,6 +321,8 @@ export class FileUseCases {
     return this.fileRepository.getFilesCountWhere({
       userId,
       status: FileStatus.TRASHED,
+      deleted: true,
```
sg-gs, line 324: Remove this, we are deprecating these fields for files

sg-gs (changes requested):

src/modules/file/file.usecase.ts
```diff
@@ -321,6 +321,7 @@ export class FileUseCases {
     return this.fileRepository.getFilesCountWhere({
       userId,
       status: FileStatus.TRASHED,
+      removed: false,
```
sg-gs, line 324: Remove this field, is going to be deprecated, **only status should be used**

src/modules/trash/trash.usecase.ts
```diff
@@ -33,10 +37,10 @@ export class TrashUseCases {
       await this.folderUseCases.deleteByUser(trashOwner, folders);
     }
 
-    for (let i = 0; i < count; i += emptyTrashChunkSize) {
+    for (let i = 0; i < filesCount; i += emptyTrashChunkSize) {
       const files = await this.fileUseCases.getFiles(
         trashOwner.id,
-        { deleted: true, removed: false },
+        { status: FileStatus.TRASHED, removed: false },
```
sg-gs, line 43: Remove the `removed` field usage, **only status should be used.**

sg-gs (changes requested):

src/modules/trash/trash.usecase.ts
miguelsw: @sg-gs  it was updated in commits after this comment

sg-gs (approved):

PR #208, opened 2023-09-12
[_]: feat/sharing-mail-notifications

sg-gs:
Send notifications to the involved users when sharing actions happen.


src/modules/sharing/sharing.service.ts
```diff
@@ -620,7 +623,26 @@ export class SharingService {
...
+      if (createInviteDto.notifyUser) {
+        new MailerService(this.configService)
+          .sendInvitationToSharingReceivedEmail(
+            user.email,
+            userJoining.email,
+            item.plainName,
+            {
+              acceptUrl: this.configService.get('clients.drive.web'),
+              declineUrl: this.configService.get('clients.drive.web'),
+              message: 'Hello, I want to share with you',
```
CandelR, line 637: forgot to add the optional message that has been received in the payload

PR #215, opened 2023-09-28
PB-722: create your own sharing when the first user accepts the invit…

edisonjpadilla:
create your own sharing when the first user accepts the invitation, we do this to keep the file shared although all the users are removed from it.

### DIFF
--- a/src/modules/sharing/sharing.service.ts
+++ b/src/modules/sharing/sharing.service.ts
@@ -47,6 +47,8 @@ import { CreateSharingDto } from './dto/create-sharing.dto';
 import { aes } from '@internxt/lib';
 import { Environment } from '@internxt/inxt-js';
 
+const SHARE_WITH = '00000000-0000-0000-0000-000000000000';
+
 export class InvalidOwnerError extends Error {
   constructor() {
     super('You are not the owner of this folder');
@@ -1115,7 +1117,7 @@ export class SharingService {
     const newSharing = Sharing.build({
       ...dto,
       id: v4(),
-      sharedWith: '00000000-0000-0000-0000-000000000000',
+      sharedWith: SHARE_WITH,
       type: SharingType.Public,
       ownerId: user.uuid,
       createdAt: new Date(),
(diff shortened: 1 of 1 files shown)


sg-gs:
Why are we doing this @edisonjpadilla ?

edisonjpadilla:
@sg-gs that was to keep this behavior https://inxt.atlassian.net/browse/PB-722?focusedCommentId=14642

edisonjpadilla:
> Why are we doing this @edisonjpadilla ?

@sg-gs  that was to keep this behavior https://inxt.atlassian.net/browse/PB-722?focusedCommentId=14642

sg-gs:
Okay, is this sharing being removed / should be removed once we click the stop sharing button? @edisonjpadilla @aboredvaro

src/modules/sharing/sharing.service.ts
```diff
@@ -1212,6 +1214,23 @@ export class SharingService {
         ? acceptInviteDto.encryptionKey
         : invite.encryptionKey,
     });
+
+    const ownerSharing = await this.sharingRepository.countSharing({
```
sg-gs, line 1218: Use a findOne from the repository, instead of a count. Also, if you are creating a sharing that is shared with the owner itself, the `sharedWith` should be the `owner.uuid`. The user with the uuid 0000... is a special Drive user which also is used as a reference for public sharing (which is not the case here, as you are accepting an invitation = private sharing).

sg-gs (changes requested):

src/modules/sharing/sharing.service.ts
edisonjpadilla: I'm using the `count` method because is more efficient and has better performance, why should I use the `findOne` method? 

 About the `sharedWith` I'll make the change.

src/modules/sharing/sharing.repository.ts
```diff
@@ -541,4 +541,8 @@ export class SequelizeSharingRepository implements SharingRepository {
       where,
     });
   }
+
+  async countSharing(where: Partial<Sharing>): Promise<number> {
```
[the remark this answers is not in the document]
edisonjpadilla: yes you're right. Let me see if the typescript doesn't say anything.

apsantiso:
Closing this. I has not been updated for months, reopen if you need.

PR #182, opened 2023-07-17
[_]: fix/e2e tests

cg27shokworks:

PR #217, opened 2023-10-17
PB-1072: fix unit tests

edisonjpadilla:
The unit tests are passing correctly.

The E2E test has been fixed in this PR: https://github.com/internxt/drive-server-wip/pull/182


edisonjpadilla:
Once we merge this PR, everything we do from now on must have tests, otherwise the PR will not be approved, right @apsantiso @sg-gs ?

edisonjpadilla:
@apsantiso  remember that we changed the version of nodejs, the tests require v16, we have to adjust it to the new version

apsantiso:
the run-test(16x) pipeline is completely commented out. I already updated the version to node 20... I don't know what do you mean with updating.

Regarding the "everything needs to have tests", it depends on the priority and urgency of the feature. Also pretty sure there are things that are just going to take too much time to be tested and 0 worth on it so we can skip them. 

So if it is new code, ideally we should try to add tests as long as we dont spend weeks just for the tests while blocking other things.

edisonjpadilla:
I think if the run-test(16x) pipeline will never work, it's better delete it

apsantiso (dismissed):

PR #218, opened 2023-10-20
[PB-995]: Feat/enable invite public sharing

apsantiso:
- Added update sharing Endpoint
- Added get sharing type endpoint
- Allow invited users to get public sharing
- Prevent error by not looking for 000-000 user's role.
- Stop removing sharings on invite and createPublicShare

### DIFF
--- a/src/modules/sharing/sharing.repository.ts
+++ b/src/modules/sharing/sharing.repository.ts
@@ -176,6 +176,22 @@ export class SequelizeSharingRepository implements SharingRepository {
     return raw ? SharingRole.build(raw) : null;
   }
 
+  async findOneByOwnerOrSharedWithItem(
+    userId: User['uuid'],
+    itemId: Sharing['itemId'],
+    itemType: Sharing['itemType'],
+  ): Promise<Sharing> {
+    const raw = await this.sharings.findOne({
+      where: {
+        itemId,
+        itemType,
+        [Op.or]: [{ ownerId: userId }, { sharedWith: userId }],
+      },
+    });
+
+    return raw ? Sharing.build(raw) : null;
+  }
+
   private async findSharingRoles(
     where: Partial<SharingRole>,
   ): Promise<SharingRole[]> {
--- a/src/modules/sharing/sharing.repository.ts
+++ b/src/modules/sharing/sharing.repository.ts
@@ -89,6 +89,9 @@ export class SequelizeSharingRepository implements SharingRepository {
     return this.findSharingsWithRoles({
       itemId: item.uuid,
       itemType: (item as any).fileId ? 'file' : 'folder',
+      sharedWith: {
+        [Op.not]: '00000000-0000-0000-0000-000000000000',
+      },
     });
   }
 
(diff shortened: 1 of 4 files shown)


src/modules/sharing/sharing.service.ts
```diff
@@ -1799,4 +1787,67 @@ export class SharingService {
...
+    const isUserOwner = item.isOwnedBy(user);
+
+    if (!isUserOwner) {
+      throw new ForbiddenException();
+    }
+
+    const oldSharingType =
+      type === SharingType.Private ? SharingType.Public : SharingType.Private;
+
+    await this.sharingRepository.updateSharing(
```
apsantiso, line 1813: Let me know if we expect to delete the public sharing when switching between modes.
I used update so the previous links can continue being accessible instead of creating a new one.

src/modules/sharing/sharing.repository.ts
```diff
@@ -283,14 +312,8 @@ export class SequelizeSharingRepository implements SharingRepository {
...
         [sequelize.literal(`MAX("SharingModel"."created_at")`), 'createdAt'],
-        [sequelize.literal(`"SharingModel"."type"`), 'sharingType'],
-      ],
-      group: [
-        'folder.id',
-        'folder->user.id',
-        'SharingModel.item_id',
-        'sharingType',
       ],
+      group: ['folder.id', 'folder->user.id', 'SharingModel.item_id'],
```
apsantiso, line 316: Please, take into account that if this is merged before the code in https://github.com/internxt/drive-web/pull/892 is deployed into production, it will break the front as they need `sharingType`. 

I can add this again so it can be merged without breaking the front, but we need to remove `sharingType ` again before moving the front code into prod.

src/modules/sharing/sharing.repository.ts
```diff
@@ -342,14 +365,8 @@ export class SequelizeSharingRepository implements SharingRepository {
...
         [sequelize.literal(`MAX("SharingModel"."created_at")`), 'createdAt'],
-        [sequelize.literal(`"SharingModel"."type"`), 'sharingType'],
-      ],
-      group: [
-        'file.id',
-        'file->user.id',
-        'SharingModel.item_id',
-        'sharingType',
       ],
+      group: ['file.id', 'file->user.id', 'SharingModel.item_id'],
```
apsantiso, line 369: Same as above.

src/modules/sharing/sharing.service.ts
```diff
@@ -1799,4 +1787,67 @@ export class SharingService {
...
+      {
+        itemId,
+        itemType,
+        sharedWith: '00000000-0000-0000-0000-000000000000',
```
sg-gs, line 1817: If this is not public, how are we going to use this as a condition in any case? Will not break when the the sharing is private?

src/modules/sharing/sharing.service.ts
```diff
@@ -1122,18 +1112,16 @@ export class SharingService {
...
-      SharingType.Private,
-    );
-
     const sharing = await this.sharingRepository.findOneSharingBy({
       itemId: dto.itemId,
       itemType: dto.itemType,
       type: SharingType.Public,
     });
 
+    if (!item.isOwnedBy(user) && !sharing) {
```
sg-gs, line 1121: Why ownership is enforced together with the sharing existence? Shouldn't we enforce it always?

sg-gs (changes requested):

src/modules/sharing/sharing.service.ts
apsantiso: If is not public, it just doesn't try to get roles or something like that for sharings with 0000-0000. 

It does not break when the sharing is private. At least not with our current implementation, which is mainly using 
`findSharingsWithRolesByItem` to fetch private users and their roles.

- We could add some flag to deactivate the link without changing it to private if we dont want the link to be invalidated between type changes. 
- Deleting it is an option too, but given that case, the previous LINK is going to be invalidated.

PR #222, opened 2023-10-31
[PB-1041, PB-1042, PB-1040]: Feat/pre create users

apsantiso:
1. Create migrations for pre created users
2. Enable pre create users endpoint.
3. Enable pre created users invitation. 
4. send email when a pre created user is invited.
5. Migrate invitations if user registers via normal signup flow.


src/modules/sharing/models/index.ts
```diff
@@ -162,6 +163,13 @@ export class SharingInviteModel
   })
   sharedWithUser: UserModel;
 
+  @BelongsTo(() => PreCreatedUserModel, {
+    foreignKey: 'sharedWith',
+    targetKey: 'uuid',
+    as: 'preCreatedUser',
+  })
+  sharedWithPreCreatedUser: UserModel;
```
sg-gs, line 171: I think should be: 
`sharedWithPreCreatedUser: PreCreatedUserModel;`
Isn't?

src/modules/sharing/sharing.controller.ts
```diff
@@ -180,6 +182,18 @@ export class SharingController {
...
 
+  @UseGuards(ThrottlerGuard)
+  @Get('/invites/:id/validate')
+  @Public()
+  @ApiParam({
+    name: 'id',
+    description: 'Id of the invite to validate',
+    type: String,
+  })
+  async validateInvite(@Param('id') id: SharingInvite['id']) {
```
sg-gs, line 193: async is not needed here, although you do a try catch. I would advise adding a try-catch with a log.

src/modules/user/pre-created-users.repository.ts
```diff
@@ -0,0 +1,44 @@
...
+
+@Injectable()
+export class SequelizePreCreatedUsersRepository {
+  constructor(
+    @InjectModel(PreCreatedUserModel)
+    private modelUser: typeof PreCreatedUserModel,
+  ) {}
+
+  async findById(id: number): Promise<PreCreatedUserModel | null> {
+    const user = await this.modelUser.findByPk(id);
```
sg-gs, line 16: return directly instead of using async - await. Also, we are not converting this to domain

src/modules/user/pre-created-users.repository.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  async findByUsername(
+    username: PreCreatedUserAttributes['username'],
+  ): Promise<PreCreatedUserModel | null> {
+    const user = await this.modelUser.findOne({
+      where: {
+        username,
+      },
+    });
+    return user;
```
sg-gs, line 36: Not converted to the domain

sg-gs (changes requested):

sg-gs:
Nice job here @apsantiso. Give to this a thought: 
- What if a bad actor tries to pre-register a given list of emails to prevent those real users to sign up? 

I suggest to handle that case on the current signup, adding a path where if the user is pre-created and tries to sign up, we remove the pre-created one and migrate the invites to the real new user. Thus, we do not allow anyone to be able to block users from joining Internxt.

apsantiso:
> Nice job here @apsantiso. Give to this a thought:
> 
> * What if a bad actor tries to pre-register a given list of emails to prevent those real users to sign up?
> 
> I suggest to handle that case on the current signup, adding a path where if the user is pre-created and tries to sign up, we remove the pre-created one and migrate the invites to the real new user. Thus, we do not allow anyone to be able to block users from joining Internxt.

@sg-gs Yes, I thought about handling that case together with the ticket related to register pre created users as we dont want to block any user from signin up. However, that tickets seems to be taking time so I don't have problem with covering that case in this PR quickly.

src/modules/user/user.usecase.ts
```diff
@@ -385,6 +405,110 @@ export class UserUseCases {
...
+    const { privateKey: encPrivateKey } = preCreatedUser;
+    const privateKey = aes.decrypt(encPrivateKey, defaultPass);
+
+    const invites = await this.sharingRepository.getInvitesBySharedwith(
+      preCreatedUser.uuid,
+    );
+
+    for (const invite of invites) {
+      const decryptedEncryptionKey = await decryptMessageWithPrivateKey({
+        encryptedMessage: Buffer.from(invite.encryptionKey, 'base64').toString(
```
apsantiso, line 431: frontend side we are using aob and btoa but they are marked as deprecated by Nodejs, so used Buffer to accomplish the same behavior.  Same as below with buffer.from(xxx, 'binary')

sg-gs (approved):

PR #242, opened 2023-12-06
Fix: AFS guest invitation email link fixed and remove invitation if it fails.

apsantiso:


src/modules/sharing/sharing.service.ts
```diff
@@ -1116,22 +1116,31 @@ export class SharingService {
...
-          // no op
-        });
+        );
+      } catch (error) {
+        Logger.error(
+          `[SHARING/GUESTUSEREMAIL] Error sending email pre created userId: ${
+            userJoining.uuid
+          }, error: ${JSON.stringify(error)}`,
+        );
+        await this.removeInvite(user, createdInvite.id);
```
apsantiso, line 1141: If the guest user invitation fails, there is no point on persisting this invitation.

src/modules/sharing/sharing.service.ts
```diff
@@ -1116,22 +1116,31 @@ export class SharingService {
...
-          // no op
-        });
+        );
+      } catch (error) {
+        Logger.error(
+          `[SHARING/GUESTUSEREMAIL] Error sending email pre created userId: ${
+            userJoining.uuid
+          }, error: ${JSON.stringify(error)}`,
+        );
+        await this.sharingRepository.deleteInvite(createdInvite);
```
apsantiso, line 1141: If the guest user invitation fails, there is no point in persisting this invitation.

CandelR (approved):
Now it works perfect 👌

sg-gs (approved):

PR #243, opened 2023-12-11
[PB-1144, PB-1143, PB-608]: feat/update-user-email

edisonjpadilla:
**The general flow for changing the email and accepting an attempt to change the email (attempt-change-email) is as follows:**

**Creation of the Email Change Attempt:**

1. A user requests to change their email address.
2. A new record is created in the `users_attempt_change_email` table containing information related to the email change attempt.
3. A verification link is generated and sent to the user's new email address.

**Sending the Verification Link:**

1. The MailerService is used to send an email to the user's new email address.
2. The email contains a verification link pointing to a route in the web application.

**Verification of the Link:**

1. When the user clicks on the verification link, the application processes the request.
2. The application verifies the validity of the link, ensuring it has not expired and has not been verified previously.

**Acceptance of the Email Change:**

1. If the link is valid, the user takes the action to accept the email change.
2. A database transaction is performed to update the user's email address.
3. The email change attempt is marked as verified.

**Updating Email in Third-Party Components:**

1. The email address is updated in other components, such as the external `bridge service (BridgeService)` and `shared space invitations`.
2. Possible errors, such as the email address already in use or the user not found, are handled.

*This flow aims to ensure security and proper verification before making changes to user information. Updates are also made to other components to ensure information consistency throughout the application.*


.env.template
```diff
@@ -27,6 +27,7 @@ SENDGRID_TEMPLATE_DRIVE_SHARING_USER_REMOVED=d-0c9cbd0a649d4cee8eac6a4ae68736c9
 SENDGRID_TEMPLATE_DRIVE_SHARING_ROLE_UPDATED=d-0c9cbd0a649d4cee8eac6a4ae68736c9
 SENDGRID_TEMPLATE_SEND_LINK_CREATE_RECEIVER=d-eb5a1fd73d764e9991a25e2a0297f279
 SENDGRID_TEMPLATE_SEND_LINK_CREATE_SENDER=d-7889146930fa421083b4bf1cdcaedab3
+SENDGRID_TEMPLATE_DRIVE_UPDATE_USER_EMAIL=d-46a66194ad8e4bb6919d39e20d9c34d1
```
edisonjpadilla, line 30: please remember to add this variable @sg-gs

src/externals/mailer/mailer.service.ts
```diff
@@ -147,4 +147,21 @@ export class MailerService {
...
+    newEmail: string,
+    encryptedAttemptId: string,
+  ) {
+    const webUrl = this.configService.get('clients.drive.web');
+
+    const verificationUrl = `${webUrl}/verify-new-email/${encryptedAttemptId}`;
+
+    await this.send(
+      newEmail,
+      this.configService.get('mailer.templates.updatedSharingRole'),
```
CandelR, line 161: forgot to change the template, it is `updateUserEmail`

migrations/20231211002759-create_users_attempt_change_email_table.js
```diff
@@ -0,0 +1,47 @@
...
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.createTable(tableName, {
+      id: {
+        type: Sequelize.INTEGER,
+        allowNull: false,
+        primaryKey: true,
+        autoIncrement: true,
+      },
+      user_id: {
```
sg-gs, line 14: relate by uuid and establish the foreign key please

src/modules/user/attempt-change-email/attempt-change-email.controller.ts
sg-gs, line 1: This is a user-related domain use case, why are we adding a module just for this? This seems like an overkill as we just need a new repo for the temp table

sg-gs (changes requested):

sg-gs:
Please add a description of the proposed changes @edisonjpadilla

src/modules/user/attempt-change-email/attempt-change-email.controller.ts
edisonjpadilla: It is not just a new repository, I created it as a submodule in UserModule because it handles new exceptions, atomic transactions, statuses, and sends mail, I also made it following the modular Nestjs structure which provides good long-term scalability

migrations/20231211002759-create_users_attempt_change_email_table.js
edisonjpadilla: I'm gonna do it, but, could you tell me why, please?

src/modules/user/attempt-change-email/attempt-change-email.controller.ts
sg-gs: How many use cases are you going to have when changing the email? This is a small feature that was blocking us due to the need for deep structural changes in the network service. The email change is not going to expand to something bigger than for example, the storage, with files, folders, and thumbnails, which englobe multiple use cases (a file system hierarchy, with aggregate calculations, interaction with the network, and the need of massive deletions). 

I find this an overkill mainly because is not going to grow (thus, no scalability issue here), fits well in the user's domain (as it is exclusively for users) and also, the mailing thing is handled by the event bus - or should be -, as the email change event could be interesting for the notifications service also - not only for the mailing -, so the desktop apps could renew their tokens on demand.

migrations/20231211002759-create_users_attempt_change_email_table.js
sg-gs: An auto-incremental ID has a well-known limit where it overflows. Part of our legacy system involves using the users.id, files.id, etc, therefore, we need to stop creating new references while removing the already existent ones to use UUIDs, which use their data type that does not have a limit apart from using all the UUIDs (5.3 x 10^36 possibilities). 

If you reach the limit, you just cannot insert new records. We are far from that, but we should take care before it's near as it would mean we will have x100 more users and the risk would be imminent.

.env.template
sg-gs: Done

sg-gs:
Hey @edisonjpadilla the tests are broken in this PR. We need the unit tests to run successfully to test and merge this with confidence.

sg-gs:
Migration run:

```

== 20231212002759-create-users-attempt-change-email-table: migrating =======
== 20231212002759-create-users-attempt-change-email-table: migrated (0.196s)

✨  Done in 2.76s.
```

src/modules/user/user.usecase.ts
```diff
@@ -808,4 +812,123 @@ export class UserUseCases {
...
+    const { uuid, email } = user;
+
+    await this.userRepository.updateById(user.id, {
+      email: newEmail,
+      username: newEmail,
+      bridgeUser: newEmail,
+    });
+    await this.sharedWorkspaceRepository.updateGuestEmail(email, newEmail);
+
+    this.networkService.updateUserEmail(uuid, newEmail).catch(async () => {
```
sg-gs, line 841: The deal was: 
1. Update the network email
2. Update the drive email references

If 2 fails -> do the rollback of 1. 

But here I see 2 before 1.

sg-gs (changes requested):

sg-gs (approved):

PR #247, opened 2023-12-15
[PB-1379] Feat:password protected sharing

apsantiso:
Users need to share their public links with a password to protect their data. To accomplish this with zero knowledge, the next changes are proposed.

- Add encryptedPassword field to the sharingModel.
- `encryptedPassword` needs to be encrypted with the public `code` (which we don't store).
- Allow frontend to create a public sharing with password in one step (just enable encryptedPassword in the DTO)
- Enable one endpoint to edit/create password for an already created public sharing. 
- Enable one endpoint to remove password protection from a public sharing.
- Enable one endpoint to return public shared item info. We need to show users basic information of the item they are going to access to.  (we just return `size`, `plainName` and `type`).


src/modules/sharing/sharing.controller.ts
```diff
@@ -71,11 +73,67 @@ export class SharingController {
...
+    name: 'sharingId',
+    description: 'Id of the sharing',
+    type: String,
+  })
+  @ApiOkResponse({ description: 'Get sharing item info' })
+  async getPublicSharingItemInfo(@Param('sharingId') sharingId: Sharing['id']) {
+    return this.sharingService.getPublicSharingItemInfo(sharingId);
+  }
+
+  @Post('/:sharingId/password')
```
sg-gs, line 99: Shouldn't this be a PATCH?

src/modules/sharing/sharing.controller.ts
```diff
@@ -71,11 +73,67 @@ export class SharingController {
...
+    description: 'Id of the sharing',
+    type: String,
+  })
+  @ApiOkResponse({ description: 'Sets/edit password for public sharings' })
+  async setPublicSharingPassword(
+    @UserDecorator() user: User,
+    @Param('sharingId') sharingId: Sharing['id'],
+    @Body() sharingPasswordDto: SetSharingPasswordDto,
+  ) {
+    const { encryptedPassword } = sharingPasswordDto;
```
sg-gs, line 114: Could we somewhat validate the format of this? Is it not an empty string, has the expected format (hex or whatever, etc)?

src/modules/sharing/sharing.service.ts
```diff
@@ -234,6 +242,17 @@ export class SharingService {
       throw new ForbiddenException();
     }
 
+    if (sharing.isProtected() && !plainPassword) {
+      throw new PasswordNeededError();
```
sg-gs, line 246: nice check

src/modules/sharing/sharing.service.ts
```diff
@@ -234,6 +242,17 @@ export class SharingService {
...
+    }
+
+    if (sharing.isProtected()) {
+      const decryptedPassword = aes.decrypt(sharing.encryptedPassword, code);
+      if (decryptedPassword !== plainPassword) {
```
sg-gs, line 251: Do not provide a message for this kind of situation, mix it with other Forbidden Exceptions, so we do not provide context to a brute-force attacker, even if we are using aes256 and a strong key, it is always a good practice to take into count.

src/modules/sharing/sharing.service.ts
```diff
@@ -286,6 +305,89 @@ export class SharingService {
...
+  ): Promise<Sharing> {
+    const sharing = await this.sharingRepository.findOneSharing({
+      id,
+    });
+
+    if (!sharing) {
+      throw new NotFoundException();
+    }
+
+    if (!sharing.isPublic() || !sharing.isOwnedBy(user)) {
```
sg-gs, line 361: If it is not public and someone attempts to set a password, instead of Forbidden, I suggest returning a ConflictException / BadRequestException as it is more semantic/explanatory.

src/modules/sharing/sharing.service.ts
```diff
@@ -286,6 +305,89 @@ export class SharingService {
...
+  async removeSharingPassword(user: User, id: Sharing['id']): Promise<Sharing> {
+    const sharing = await this.sharingRepository.findOneSharing({
+      id,
+    });
+
+    if (!sharing) {
+      throw new NotFoundException();
+    }
+
+    if (!sharing.isPublic() || !sharing.isOwnedBy(user)) {
```
sg-gs, line 381: Same here

sg-gs (changes requested):

src/modules/sharing/sharing.controller.ts
apsantiso: The Dto already throws an error if the string is empty with  decorator @IsNotEmpty(). 

The password is limited to 50 characters in the frontend, but we receive it already encrypted so we can not check this.

However, I see `@internxt/lib` encodes the returned encryptedText into base64, so I added the @IsBase64() check to `set-sharing-password.dto.ts`. What do you think?

sg-gs (dismissed):
Please change the labels of the tests to accommodate 'When, then' pattern. Apart from that is all OK.

src/main.ts
```diff
@@ -22,17 +22,7 @@ async function bootstrap() {
...
-        'Content-Type',
-        'Authorization',
-        'method',
-        'internxt-version',
-        'internxt-client',
-        'internxt-mnemonic',
-        'x-share-password',
-        'X-Internxt-Captcha',
-      ],
+      allowedHeaders: ['*'],
```
sg-gs, line 25: Why? This a security issue

sg-gs (changes requested):

sg-gs:
```

== 20231211135328-add-encrypted-password-field-to-sharings: migrating =======
== 20231211135328-add-encrypted-password-field-to-sharings: migrated (0.170s)

✨  Done in 2.96s.
```

sg-gs (approved):

PR #248, opened 2023-12-19
[_]: fix/e2e tests pipeline

apsantiso:
This pull requests is just for separating the previous PR made by Edison. 

It also intends to fix the production build problem that was showing when e2e tests are fixed.

How to test build for production:

1. Clone your .env.development and change its name to .env (to simulate production .env file)
2. Run docker run --rm -it $(docker build -q .) in the root folder of the project.
3. Database connection is going to fail but the container should start without any extra error.


sg-gs:
If it is a WIP @apsantiso, keep the PR as a Draft, once ready, change the status to Ready for Review

PR #250, opened 2024-01-02
[PB-1402]: feat/auto account unblock

apsantiso:
Users need to be able to auto unblock their account. Tokens must have an expiration date of 48h, and everytime a new link is issued, we need to "expire" older ones.  

This PR contains:

- JWT are stateless, so we use `lastMailSent` date in `mail_limits` table to expire tokens issued before that date.
- We expire all tokens issued before lastPasswordChangedAt date, this includes session tokens and auto account unblock jwt tokens.
- If the token is not expired and is correct, then we proceed to update users'  `lastPasswordChangedAt` and reset their login attemps.

Mail Limits:
- We need to set diary/any frame of time mail limits. There was a previous implemententation on drive-server, so this PR ports that to this codebase.
- Added migration to enable unblock account mail type in `mail_limits`.


src/modules/user/user.controller.ts
```diff
@@ -428,6 +430,98 @@ export class UserController {
...
+    }
+
+    if (
+      !decodedContent.payload ||
+      !decodedContent.payload.action ||
+      !decodedContent.payload.uuid ||
+      !decodedContent.payload.email ||
+      decodedContent.payload.action !== AccountTokenAction.Unblock ||
+      !validate(decodedContent.payload.uuid)
+    ) {
```
apsantiso, line 497: Followed same pattern as the recovery flow, however, we can move the token verification to the useCase if needed.

.env.template
```diff
@@ -28,6 +28,7 @@ SENDGRID_TEMPLATE_DRIVE_SHARING_ROLE_UPDATED=d-0c9cbd0a649d4cee8eac6a4ae68736c9
 SENDGRID_TEMPLATE_SEND_LINK_CREATE_RECEIVER=d-eb5a1fd73d764e9991a25e2a0297f279
 SENDGRID_TEMPLATE_SEND_LINK_CREATE_SENDER=d-7889146930fa421083b4bf1cdcaedab3
 SENDGRID_TEMPLATE_DRIVE_UPDATE_USER_EMAIL=d-46a66194ad8e4bb6919d39e20d9c34d1
+SENDGRID_TEMPLATE_DRIVE_UNBLOCK_ACCOUNT=d-d91905dc7a7549ada00d0fc0bb0a55b1
```
apsantiso, line 31: This is the actual template ID. You can copy it before merging.

sg-gs:
We should avoid adding new columns to the user's table unless it is a core deep-related user attribute. This table is queried a lot and adding new columns affects to the database, which stores rows in order. 

Let's work on adding new tables/relations to the user's table, through their UUID. Also, we can use the timestamp `lastPasswordChangedAt` to handle this situation, because in the end, is the same thing. Then, we check that value on the middleware, so if the password is older than the `lastPasswordChangedAt`, we consider it as 'expired'. There is no need to store the entire token and do a hard match, that is an overkill in memory without any advantage. @apsantiso 

Let's close with @larry-internxt this one: https://github.com/internxt/drive-server-wip/pull/148/files

src/modules/user/dto/account-unblock.dto.ts
```diff
@@ -0,0 +1,11 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { IsNotEmpty } from 'class-validator';
+
+export class RequestAccountUnblock {
```
sg-gs, line 4: Are we able to validate the email format?

sg-gs (changes requested):

src/lib/jwt.ts
```diff
@@ -47,5 +47,5 @@ export function getTokenDefaultIat() {
 }
 
 export function isTokenIatGreaterThanDate(date: Date, iat: number) {
-  return Math.floor(date.getTime() / 1000) < iat;
+  return Math.floor(date.getTime() / 1000) <= iat;
```
apsantiso, line 50: Not considering the same date as valid is a bit troublesome. This will cause problems also with the "password change".

src/modules/user/user.usecase.ts
```diff
@@ -698,6 +700,72 @@ export class UserUseCases {
...
+    if (!user) {
+      throw new NotFoundException();
+    }
+
+    const defaultIat = getTokenDefaultIat();
+
+    await this.userRepository.updateByUuid(user.uuid, {
+      lastPasswordChangedAt: new Date(defaultIat * 1000),
+    });
+
```
apsantiso, line 716: I used defaultIat as reference because it is possible to lose some miliseconds when we round a number for the Iat. Thus, it is better to use it as reference to create the date.

src/modules/user/user.usecase.ts
```diff
@@ -698,6 +700,72 @@ export class UserUseCases {
...
+        new Date(user?.lastPasswordChangedAt),
+        tokenIat,
+      )
+    ) {
+      throw new ForbiddenException();
+    }
+
+    await this.userRepository.updateByUuid(userUuid, {
+      errorLoginCount: 0,
+      lastPasswordChangedAt: null,
```
apsantiso, line 765: @sg-gs Do you recommend to set to null the lastPasswordChangedAt field or we can just leave it?

sg-gs:
Secret `SENDGRID_TEMPLATE_DRIVE_UNBLOCK_ACCOUNT` added in production.

src/lib/jwt.ts
sg-gs: If is a bit troublesome why are we doing it?

src/modules/user/user.controller.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+  describe('POST /unblock-account', () => {
+    it('When user is not found, then returns NotFoundException', async () => {
+      userUseCases.sendAccountUnblockEmail.mockRejectedValueOnce(
+        new NotFoundException(),
+      );
+      await expect(
+        userController.requestAccountUnblock({ email: 'test@test.com' }),
+      ).rejects.toThrow(NotFoundException);
+    });
+    it('When an unexpected error is throw, then returns InternalServerException', async () => {
```
sg-gs, line 67: _When an unexpected error is thrown, then the error is propagated_ could be more descriptive, especially to avoid relating legend with explicit code

src/modules/user/user.controller.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+      sharingService,
+    );
+  });
+
+  it('should be defined', () => {
+    expect(userController).toBeDefined();
+  });
+
+  describe('POST /unblock-account', () => {
+    it('When user is not found, then returns NotFoundException', async () => {
```
sg-gs, line 59: Let's talk in HTTP Status Codes better, 404 for instance, as we would like to avoid relating the test with literal code as if it changes, we need to change this also.

src/modules/user/user.controller.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+        },
+        getEnv().secrets.jwt,
+        '-48h',
+      );
+      await expect(userController.accountUnblock(expiredToken)).rejects.toThrow(
+        ForbiddenException,
+      );
+    });
+
+    it('When token is valid but useCase throws badRequest or Forbidden, then fails with respective error', async () => {
```
sg-gs, line 131: Why is this important? I mean, wouldn't this be a unique test that tells us that the errors are propagated instead of one test per each type of error?

src/modules/user/user.controller.ts
```diff
@@ -428,6 +430,97 @@ export class UserController {
...
+  @UseGuards(ThrottlerGuard)
+  @Post('/unblock-account')
+  @HttpCode(200)
+  @ApiOperation({
+    summary: 'Request account unblock',
+  })
+  @Public()
+  async requestAccountUnblock(@Body() body: RequestAccountUnblock) {
+    try {
+      return await this.userUseCases.sendAccountUnblockEmail(body.email);
```
sg-gs, line 442: `return` directly or assign to variable and `return` the constant

src/modules/user/user.usecase.spec.ts
```diff
@@ -249,6 +266,105 @@ describe('User use cases', () => {
...
+        jest.setSystemTime(fixedSystemCurrentDate);
+      });
+
+      afterAll(async () => {
+        jest.useRealTimers();
+      });
+
+      it('When user does not exist, then fail', async () => {
+        const userFindByEmailSpy = jest.spyOn(userRepository, 'findByEmail');
+        const email = 'email@test.com';
```
sg-gs, line 284: Use data generators if possible, please

src/modules/user/user.usecase.spec.ts
```diff
@@ -249,6 +266,105 @@ describe('User use cases', () => {
...
+        userFindByEmailSpy.mockReturnValueOnce(null);
+
+        await expect(
+          userUseCases.sendAccountUnblockEmail(email),
+        ).rejects.toThrow(NotFoundException);
+      });
+
+      it('When user user exists, then user lastPasswordChangedAt is updated', async () => {
```
sg-gs, line 292: _When the user exists_, typo here?

src/modules/user/user.usecase.spec.ts
```diff
@@ -249,6 +266,105 @@ describe('User use cases', () => {
...
+      it('When user does not exist, then fail', async () => {
+        const userFindByUuidSpy = jest.spyOn(userRepository, 'findByUuid');
+        userFindByUuidSpy.mockReturnValueOnce(null);
+
+        await expect(userUseCases.unblockAccount(user.uuid)).rejects.toThrow(
+          BadRequestException,
+        );
+      });
+
+      it('When token is older than lastPasswordChangedAt, then fail', async () => {
```
sg-gs, line 330: Is not older, is previous to the date, I find this legend misleading as 'older' often means after when talking about time

src/modules/user/user.usecase.ts
```diff
@@ -698,6 +700,72 @@ export class UserUseCases {
     });
   }
 
+  async sendAccountUnblockEmail(email: User['email']): Promise<void> {
+    const secret = this.configService.get('secrets.jwt');
+    const user = await this.userRepository.findByEmail(email);
+
+    if (!user) {
+      throw new NotFoundException();
```
sg-gs, line 708: Let's `return` here, as it is more privacy-respectful.

src/modules/user/user.usecase.ts
```diff
@@ -698,6 +700,72 @@ export class UserUseCases {
...
+          action: AccountTokenAction.Unblock,
+        },
+        iat: defaultIat,
+      },
+      secret,
+      '48h',
+    );
+
+    const driveWebUrl = this.configService.get('clients.drive.web');
+    const unblockAccountTemplateId = this.configService.get(
```
sg-gs, line 731: We need to use mail limits here (10 per day is fine), as this is exploitable in terms of sending requests to auto-unblock your account thus, spending our mails per day on our mailing provider and effectively leaving us without mails for the rest of the customers which would need that.

src/modules/user/user.usecase.ts
sg-gs: We need to expire other sessions for security reasons

sg-gs (changes requested):

src/modules/user/user.controller.ts
apsantiso: If we return without the await (I know it looks odd) the catch in the controller can not get the error so I will asssign it to a variable.

src/lib/time.ts
```diff
@@ -31,4 +31,18 @@ export class Time {
   public static resumeTime(): void {
     Time.freeze = null;
   }
+
+  public static isToday = (date: Date) => {
+    const todayDate = new Date();
+
+    if (
```
sg-gs, line 38: return the comparisons directly

src/modules/user/user.controller.ts
```diff
@@ -428,6 +431,100 @@ export class UserController {
...
+    try {
+      const decoded = verifyWithDefaultSecret(token);
+      if (typeof decoded === 'string') {
+        throw new ForbiddenException();
+      }
+      decodedContent = decoded as {
+        payload: { uuid: string; action: string; email: string };
+        iat: number;
+      };
+    } catch (err) {
```
sg-gs, line 486: If you are sure about the error failure, just `catch {` without the `err`

sg-gs (changes requested):

src/modules/user/user.usecase.ts
```diff
@@ -698,6 +712,84 @@ export class UserUseCases {
...
+    );
+
+    const tokenIssuedAtDate = Time.convertTimestampToDate(tokenIat);
+
+    if (
+      mailLimit.lastMailSent > tokenIssuedAtDate ||
+      user.lastPasswordChangedAt > tokenIssuedAtDate
+    ) {
+      throw new ForbiddenException();
+    }
```
apsantiso, line 785: I decided to not consider valid tokens if they were issued before `lastPasswordChangedAt`  basically to expire the JWT once it has been used. 

Otherwise, you can keep using the same JWT as long as it's within 48 hours and the `lastMailSent` <= token's issued date.

sg-gs (approved):

sg-gs:
Migration run successfully in 100ms. Ready to deploy @apsantiso

PR #253, opened 2024-01-03
Chore: remove middleware logger for production environment

edisonjpadilla:


src/main.ts
```diff
@@ -50,7 +50,7 @@ async function bootstrap() {
   app.use(helmet());
   app.use(apiMetrics());
 
-  app.use(RequestLoggerMiddleware);
+  // app.use(RequestLoggerMiddleware);
```
sg-gs, line 53: Use this when the environment is not production

sg-gs (changes requested):

PR #254, opened 2024-01-03
Chore: remove middleware logger for production environment

edisonjpadilla:


src/main.ts
```diff
@@ -4,6 +4,7 @@ dotenv.config({ path: `.env.${process.env.NODE_ENV}` });
 
 import { ValidationPipe, Logger } from '@nestjs/common';
 import { NestFactory, Reflector } from '@nestjs/core';
+import * as process from 'process';
```
sg-gs, line 7: Nest does manage the environment, use ConfigService (or configuration as fallback) to get the environment values.

sg-gs (changes requested):

src/main.ts
edisonjpadilla: I'm using the `configuration` to get the environment and validate when using the middleware logger

sg-gs: It appears as new line on this PR, how is that?

edisonjpadilla: because I did the `configuration` import below it, but anyway I removed it, I'm already only using the `configuration`

apsantiso (approved):

sg-gs (approved):

PR #255, opened 2024-01-04
PB-1142: Authenticate the user with the new email

edisonjpadilla:


src/modules/user/user.usecase.ts
```diff
@@ -932,6 +932,16 @@ export class UserUseCases {
...
-    return emails;
+    const user = await this.userRepository.findByEmail(emails.newEmail);
+    const newTokenPayload = this.getNewTokenPayload(user);
+
+    return {
+      ...emails,
+      newAuthentication: {
+        token: SignEmail(user.email, this.configService.get('secrets.jwt')),
+        newToken: Sign(newTokenPayload, this.configService.get('secrets.jwt')),
+        user,
```
CandelR, line 943: We need exact same values in user object as drive-server.

CandelR (changes requested):

src/modules/user/user.usecase.ts
apsantiso: I was checking this just for curiosity and it seems that token and newToken have the same payload as in driver-server. 

However, here is the link to the same code in drive-server so Edison can check this:

https://github.com/internxt/drive-server/blob/d3308444ad9744f3ef78032848770ba5ca8c76f4/src/app/routes/auth.ts#L224

CandelR: The token and newToken are fine, that's why I'm referring to the `user` object, which is not returned in the same format. Anyway, we've already agreed on how to update it for the time being so as not to hurt the performance of the call.

apsantiso: Kk, sorry, I thought you were talking about the tokens!

sg-gs (approved):
LGTM. Wdyt @apsantiso @CandelR ?

apsantiso (approved):

CandelR (approved):

PR #256, opened 2024-01-09
[PB-267]: feat/add last passwordChangedAt middleware

apsantiso:


apsantiso:
@larry-internxt Check it and feel free to give any feedback. It is almost a copy of the middleware in https://github.com/internxt/drive-server/pull/382. Just changed a bit of the conditions so we can early exit the function.

PR #259, opened 2024-01-11
[PB-845]: feat/calculate-folder-size

edisonjpadilla:
I'm using a recursive raw query to get each file of each folder belonging to folder father. 

the properties that I used to filter in the query all have indexes;  `file.folder_uuid`, `folder.uuid`, `folder.parent_uuid,` so improving the performance for each request.

### DIFF
--- a/src/modules/folder/folder.controller.spec.ts
+++ b/src/modules/folder/folder.controller.spec.ts
@@ -1,0 +1,50 @@
+import { createMock } from '@golevelup/ts-jest';
+import { Test, TestingModule } from '@nestjs/testing';
+import { newFolder } from '../../../test/fixtures';
+import { FileUseCases } from '../file/file.usecase';
+import { FolderController } from './folder.controller';
+import { Folder } from './folder.domain';
+import { FolderUseCases } from './folder.usecase';
+import { CalculateFolderSizeTimeoutException } from './exception/calculate-folder-size-timeout.exception';
+
+describe('FolderController', () => {
+  let folderController: FolderController;
+  let folderUseCases: FolderUseCases;
+  let folder: Folder;
+
+  beforeEach(async () => {
+    const module: TestingModule = await Test.createTestingModule({
+      controllers: [FolderController],
+      providers: [
+        { provide: FolderUseCases, useValue: createMock() },
+        { provide: FileUseCases, useValue: createMock() },
+      ],
+    }).compile();
+
+    folderController = module.get<FolderController>(FolderController);
+    folderUseCases = module.get<FolderUseCases>(FolderUseCases);
+    folder = newFolder();
+  });
+
+  describe('get folder size', () => {
+    it('When get folder size is requested, then return the folder size', async () => {
+      const expectedSize = 100;
+      jest
+        .spyOn(folderUseCases, 'getFolderSizeByUuid')
+        .mockResolvedValue(expectedSize);
+
+      const result = await folderController.getFolderSize(folder.uuid);
+      expect(result).toEqual({ size: expectedSize });
+    });
+
+    it('When get folder size times out, then throw an exception', async () => {
+      jest
+        .spyOn(folderUseCases, 'getFolderSizeByUuid')
+        .mockRejectedValue(new CalculateFolderSizeTimeoutException());
+
+      await expect(folderController.getFolderSize(folder.uuid)).rejects.toThrow(
+        CalculateFolderSizeTimeoutException,
+      );
+    });
+  });
+});
--- a/src/modules/folder/folder.repository.spec.ts
+++ b/src/modules/folder/folder.repository.spec.ts
@@ -1,0 +1,97 @@
+import { createMock } from '@golevelup/ts-jest';
+import { CalculateFolderSizeTimeoutException } from './exception/calculate-folder-size-timeout.exception';
+import { SequelizeFolderRepository } from './folder.repository';
+import { FolderModel } from './folder.model';
+import { Folder } from './folder.domain';
+import { newFolder } from '../../../test/fixtures';
+
+jest.mock('./folder.model', () => ({
+  FolderModel: {
+    sequelize: {
+      query: jest.fn(() => Promise.resolve([[{ totalsize: 100 }]])),
+    },
+  },
+}));
+
+describe('SequelizeFolderRepository', () => {
+  const TIMEOUT_ERROR_CODE = '57014';
+
+  let repository: SequelizeFolderRepository;
+  let folderModel: typeof FolderModel;
+  let folder: Folder;
+
+  beforeEach(async () => {
+    folderModel = createMock<typeof FolderModel>();
+
+    repository = new SequelizeFolderRepository(folderModel);
+
+    folder = newFolder();
+  });
+
+  describe('calculate folder size', () => {
+    it('When calculate folder size is requested, then it works', async () => {
+      const calculateSizeQuery = `
+      WITH RECURSIVE folder_recursive AS (
+        SELECT 
+            fl1.uuid,
+            fl1.parent_uuid,
+            f1.size AS filesize,
+            1 AS row_num,
+            fl1.user_id as owner_id
+        FROM folders fl1
+        LEFT JOIN files f1 ON f1.folder_uuid = fl1.uuid
+        WHERE fl1.uuid = :folderUuid
+        AND fl1.removed = FALSE 
+        AND fl1.deleted = FALSE
+        AND f1.status != 'DELETED'
+        
+        UNION ALL
+        
+        SELECT 
+            fl2.uuid,
+            fl2.parent_uuid,
+            f2.size AS filesize,
+            fr.row_num + 1,
+            fr.owner_id
+        FROM folders fl2
+        INNER JOIN files f2 ON f2.folder_uuid = fl2.uuid
+        INNER JOIN folder_recursive fr ON fr.uuid = fl2.parent_uuid
+        WHERE fr.row_num < 100000
+        AND fl2.user_id = fr.owner_id
+        AND fl2.removed = FALSE 
+        AND fl2.deleted = FALSE
+        AND f2.status != 'DELETED'
+    ) 
+    SELECT COALESCE(SUM(filesize), 0) AS totalsize FROM folder_recursive;
+      `;
+
+      jest
+        .spyOn(FolderModel.sequelize, 'query')
+        .mockResolvedValue([[{ totalsize: 100 }]] as any);
+
+      const size = await repository.calculateFolderSize(folder.uuid);
+
+      expect(size).toBeGreaterThanOrEqual(0);
+      expect(FolderModel.sequelize.query).toHaveBeenCalledWith(
+        calculateSizeQuery,
+        {
+          replacements: {
+            folderUuid: folder.uuid,
+          },
+        },
+      );
+    });
+
+    it('When the folder size calculation times out, then throw an exception', async () => {
+      jest.spyOn(FolderModel.sequelize, 'query').mockRejectedValue({
+        original: {
+          code: TIMEOUT_ERROR_CODE,
+        },
+      });
+
+      await expect(repository.calculateFolderSize(folder.uuid)).rejects.toThrow(
+        CalculateFolderSizeTimeoutException,
+      );
+    });
+  });
+});
--- a/src/modules/folder/folder.repository.ts
+++ b/src/modules/folder/folder.repository.ts
@@ -6,12 +6,12 @@ import { v4 } from 'uuid';
 import { Folder } from './folder.domain';
 import { FolderAttributes } from './folder.attributes';
 
-import { UserModel } from '../user/user.model';
 import { User } from '../user/user.domain';
 import { UserAttributes } from '../user/user.attributes';
 import { Pagination } from '../../lib/pagination';
 import { FolderModel } from './folder.model';
 import { SharingModel } from '../sharing/models';
+import { CalculateFolderSizeTimeoutException } from './exception/calculate-folder-size-timeout.exception';
 
 function mapSnakeCaseToCamelCase(data) {
   const camelCasedObject = {};
@@ -59,6 +59,7 @@ export interface FolderRepository {
   ): Promise<void>;
   deleteById(folderId: FolderAttributes['id']): Promise<void>;
   clearOrphansFolders(userId: FolderAttributes['userId']): Promise<number>;
+  calculateFolderSize(folderUuid: string): Promise<number>;
 }
 
 @Injectable()
@@ -66,8 +67,6 @@ export class SequelizeFolderRepository implements FolderRepository {
   constructor(
     @InjectModel(FolderModel)
     private folderModel: typeof FolderModel,
-    @InjectModel(UserModel)
-    private userModel: typeof UserModel,
   ) {}
 
   async findAllCursor(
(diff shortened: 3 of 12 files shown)


sg-gs:
Hey @edisonjpadilla, we can not list "all" the folders/files in a given folder. That does not scale well, as with more folders/files, the longer the query. 

I do not see any abort option in case the time to make the `sum` is more than X seconds. This makes the proposed changes a security issue. Use the recursive query to traverse the tree and calculate a sum _there_. The database should do that. Also, add a timeout control to avoid long lasting queries to keep the database busy all the time.

sg-gs:
And please, provide a description of the PR + the testing coverage for the new code >80%

edisonjpadilla:
> Hey @edisonjpadilla, we can not list "all" the folders/files in a given folder. That does not scale well, as with more folders/files, the longer the query.
> 
> I do not see any abort option in case the time to make the `sum` is more than X seconds. This makes the proposed changes a security issue. Use the recursive query to traverse the tree and calculate a sum _there_. The database should do that. Also, add a timeout control to avoid long lasting queries to keep the database busy all the time.

I'm just getting the folders uuids and sum file sizes by folders uuids, I thought to do the sum in the recursive query, but I didn't because of the readability of the code and the raw query.

TO DO: 
1. Do the sum in the recursive query
2. Add a timeout in case it's taking more time than expected

src/modules/folder/folder.repository.ts
```diff
@@ -447,6 +453,25 @@ export class SequelizeFolderRepository implements FolderRepository {
     return folders.map((folder) => this.toDomain(folder));
   }
 
+  async calculateFolderSize(
+    folderUuid: string,
+    timeoutSeconds = 15,
+  ): Promise<number> {
+    const calculateSizeQuery = `SELECT calculate_folder_size('${folderUuid}') AS totalSize;`;
```
sg-gs, line 460: Use Sequelize variable interpolation here to avoid SQL Injections

```
const calculateSizeQuery = 'SELECT calculate_folder_size(:folderUuid) AS totalSize';

FolderModel.sequelize.query(calculateSizeQuery, {
  replacements: { folderUuid: folderUuid },
  type: QueryTypes.SELECT
})
```

src/modules/folder/folder.repository.ts
```diff
@@ -447,6 +453,25 @@ export class SequelizeFolderRepository implements FolderRepository {
     return folders.map((folder) => this.toDomain(folder));
   }
 
+  async calculateFolderSize(
```
sg-gs, line 456: The timeout is here, nice then! I will just use the 15-second timeout inside instead of offering a param for that, as it is a matter of the Infra and in fact, a concern of it (a concern of the repo, and not of some use case - business logic). 

But this code is not aborting the query on the database, is just resolving before, but the query will continue running, thus, the possibility of attacking the database and sending requests to keep it busy is still possible. 

Please abort the query operation when the timeout expires. Also, did we agree on that or rather than it, to return the size calculated until X items / X size?

src/modules/folder/folder.usecase.spec.ts
```diff
@@ -492,4 +493,32 @@ describe('FolderUseCases', () => {
...
+      expect(result).toBe(mockSize);
+      expect(folderRepository.calculateFolderSize).toHaveBeenCalledTimes(1);
+      expect(folderRepository.calculateFolderSize).toHaveBeenCalledWith(
+        'an-uuid',
+      );
+    });
+
+    it('should throw calculate folder size timeout exception', async () => {
+      jest
+        .spyOn(folderRepository, 'calculateFolderSize')
```
sg-gs, line 516: Why not mocking the repo to return after more time than the timeout seconds? Then, you will be testing the abort behavior

sg-gs (changes requested):

src/lib/add-timeout-to-promise.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  it('should reject with a timeout error', async () => {
+    await expect(addTimeoutToPromise(new Promise(() => {}), 1)).rejects.toThrow(
+      'Operation timed out',
+    );
+  });
+
+  it('should reject with a custom exception', async () => {
+    const customException = new Error('Custom Exception');
+    await expect(
+      addTimeoutToPromise(new Promise(() => {}), 1, customException),
```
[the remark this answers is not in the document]
edisonjpadilla: This function was removed.

src/modules/folder/folder.repository.spec.ts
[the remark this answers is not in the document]
edisonjpadilla: I removed them

src/modules/folder/folder.repository.ts
edisonjpadilla: I fixed this using the postgres timeout in the query ( SET LOCAL statement_timeout to :timeoutValue )

src/lib/add-timeout-to-promise.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param {Promise} promise - The Promise to add a timeout to.
+ * @param {number} timeoutMillis - Timeout duration in milliseconds.
+ * @param customException
+ * @returns {Promise} - A Promise that resolves with the original result or rejects with a timeout error.
+ */
+export function addTimeoutToPromise<T = any>(
+  promise: Promise<T>,
+  timeoutMillis: number,
+  customException?: Error,
+): Promise<any> {
```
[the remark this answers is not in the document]
edisonjpadilla: I removed that function, I'm now handling the timeout error with PostgreSQL using ` SET LOCAL statement_timeout TO :timeout-value`

migrations/20240112041245_create-calculate-folder-size-function.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const functionName = 'calculate_folder_size';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  up: async (queryInterface) => {
+    await queryInterface.sequelize.query(`
+        CREATE OR REPLACE FUNCTION ${functionName}(folder_uuid_param UUID) RETURNS BIGINT AS $$
```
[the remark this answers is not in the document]
edisonjpadilla: Here are the advantages of using this function instead of a recursive query:

Clarity and Maintainability:
Procedural code can be easier to understand and maintain for some developers, especially those familiar with procedural programming languages like PL/pgSQL.
The use of control flow statements (such as loops and conditionals) can make the code clearer compared to recursive SQL queries, which might be harder to grasp for some.

Performance:
In some cases, stored functions can be more efficient in terms of performance than recursive SQL queries, particularly when dealing with complex operations or large datasets.
The stored function may perform specific optimizations or access indexes more effectively.

Flexibility:
Using PL/pgSQL provides flexibility to perform more advanced operations within the function, such as handling exceptions, executing custom business logic, etc.


My conclusion: 

I understand well the benefits that you mentioned above, but taking into account this use case that we only need the size of the files that belong to each folder, we need to have more freedom, by freedom I mean only taking the properties that we are going to need, make the smallest number of queries and joins possible and target the fields that have indexes, that freedom is offered to us by PLSQL and not CTE.

In this file, I leave you the queries in both PSQL and CTE.

I would like you to backup the production database to test the queries and see which one is better.

[queries.txt](https://github.com/internxt/drive-server-wip/files/13930048/queries.txt)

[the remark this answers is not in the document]
edisonjpadilla: great, you're completly right!

src/modules/folder/folder.controller.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+      const expectedSize = 100;
+      jest
+        .spyOn(folderUseCases, 'getFolderSizeByUuid')
+        .mockResolvedValue(expectedSize);
+
+      const result = await folderController.getFolderSize(folderUuid);
+      expect(result).toEqual({ size: expectedSize });
+    });
+
+    it('When getFolderSizeController throws timeout exception, Then throws CalculateFolderSizeTimeoutException instance', async () => {
```
sg-gs, line 37: Remember the thing about the test legends:
_When an exception happens,  then it is propagated_

src/modules/folder/folder.controller.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+      const folderUuid = 'valid-uuid';
```
sg-gs, line 38: Use the [fixtures](https://github.com/internxt/drive-server-wip/blob/master/test/fixtures.ts) for these things, so you get and reuse: 
- Random data generators which have testing coverage
- A more realistic test data
- Less mantainance at virtually 0 effort cost

src/modules/folder/folder.repository.ts
```diff
@@ -447,6 +449,59 @@ export class SequelizeFolderRepository implements FolderRepository {
...
+    folderUuid: string,
+    timeoutSeconds = 15,
+  ): Promise<number> {
+    try {
+      const calculateSizeQuery = `
+      BEGIN;
```
sg-gs, line 458: `BEGIN` and `COMMIT` usage create a transaction, we want to avoid that at all costs, specially when working with frequently accessed tables like files and folders

src/modules/folder/folder.repository.ts
```diff
@@ -459,4 +514,6 @@ export class SequelizeFolderRepository implements FolderRepository {
     return domain.toJSON();
   }
 }
+
+// QUESTION: Why is it exporting from here?
```
sg-gs, line 518: Remove this comment. Remove the export if it should not be here. What do you think about this export @apsantiso ?

src/modules/folder/folder.usecase.spec.ts
```diff
@@ -492,4 +493,32 @@ describe('FolderUseCases', () => {
...
+      const result = await service.getFolderSizeByUuid('an-uuid');
+
+      expect(result).toBe(mockSize);
+      expect(folderRepository.calculateFolderSize).toHaveBeenCalledTimes(1);
+      expect(folderRepository.calculateFolderSize).toHaveBeenCalledWith(
+        'an-uuid',
+      );
+    });
+
+    it('When getFolderSizeByUuid times out, Then throw CalculateFolderSizeTimeoutException', async () => {
```
sg-gs, line 514: _When the folder size calculation times out, then throw an exception_

src/modules/folder/folder.usecase.spec.ts
```diff
@@ -492,4 +493,32 @@ describe('FolderUseCases', () => {
       }
     });
   });
+
+  describe('get folder size', () => {
+    it('When getFolderSizeByUuid is called, Then return the folder size', async () => {
```
sg-gs, line 498: _When the folder size is requested to be calculated, then it works_

src/modules/folder/folder.repository.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    repository = new SequelizeFolderRepository(folderModel);
+  });
+
+  describe('calculateFolderSize', () => {
+    const folderUuid = 'your-folder-uuid';
+    const timeoutSeconds = 15;
+
+    it('When calculateFolderSize is called, Then call the query correctly', async () => {
+      const calculateSizeQuery = `
```
sg-gs, line 29: I do not know if this matching is beneficial or not due to the maintenance implied. What do you think @edisonjpadilla @apsantiso ?

sg-gs (changes requested):

src/modules/folder/folder.repository.ts
edisonjpadilla: Yes, I know, but to be able to use a LOCAL statement_timeout, the query should be in a transaction otherwise it doesn't work.

src/modules/folder/folder.repository.spec.ts
edisonjpadilla: By mistake, we could modify the query and since it is of type string it wouldn't give us a syntax error or something like that. The test would be there to validate that the query is being built correctly.

apsantiso: Personally, I would not add this kind of test. However, I see not harm in having it. The one things I would remove is the jest.mock above, the folderModel is already Mocked with createMock() isn´t it? If it is necessary then forget what I said. @sg-gs

src/modules/folder/folder.repository.ts
edisonjpadilla: I removed the FolderModel exports from folder.repository

src/modules/folder/folder.repository.ts
edisonjpadilla: Here is the warning.

Screenshot 2024-01-16 at 11 15 19 AM

sg-gs: We can not use transactions, that is going to impact the service performance badly in the short and medium term. If the timeout is not possible without transactions, let's set a limit on the number of files to fetch in total to 100,000, so the sum is significant enough.

src/modules/folder/folder.usecase.spec.ts
sg-gs: Stubbing the return and expecting to receive the error is testing the mocking library more than any other thing. I recommend to make the sequelize model last more than the timeout limit so the timeout will trigger itself

src/modules/folder/folder.repository.spec.ts
sg-gs: I mean, the maintenance for this is hard, as @edisonjpadilla says, if we change something in the codebase, we do not have anything to alert us. Should we export the string as a constant to use it here and there? Maybe is a good workaround.

sg-gs (changes requested):

src/modules/folder/folder.repository.ts
edisonjpadilla: I can change the timeout of the global statement and when the query finishes restore it as it was before, so that it doesn't require a transaction. Please confirm which of the two ways do you implement?

edisonjpadilla: I already modified the query, what I did was create a new field as row_num in the first query and in the recusivity I was adding each iteration and put the limit with a `where statement`, something like this: `WHERE fr.row_num < 100000` , and left the timeout error handler.

edisonjpadilla:
@sg-gs Why should we test the sequelize library?  it has already been tested by the owners, we just should mocking the response and send the same response as is at the sequelize doc and test our code by that.

https://github.com/internxt/drive-server-wip/pull/259#discussion_r1453704592

sg-gs (changes requested):

src/modules/folder/folder.repository.ts
```diff
@@ -447,6 +446,53 @@ export class SequelizeFolderRepository implements FolderRepository {
...
+        UNION ALL
+        
+        SELECT 
+            fl2.uuid,
+            fl2.parent_uuid,
+            f2.size AS filesize,
+            fr.row_num + 1
+        FROM folders fl2
+        INNER JOIN files f2 ON f2.folder_uuid = fl2.uuid
+        INNER JOIN folder_recursive fr ON fr.uuid = fl2.parent_uuid
```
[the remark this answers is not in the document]
edisonjpadilla: do you refer to INNER JOIN OR WHERE STATEMENT by user_id?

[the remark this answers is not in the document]
sg-gs: I am referring to this comment I left yesterday which does the most optimum query I found: https://github.com/internxt/drive-server-wip/pull/259#discussion_r1453276292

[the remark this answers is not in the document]
edisonjpadilla: but where will I get the userId from? of the request token? or is it from the same `select`?

PR #261, opened 2024-01-12
fix: handle workspace users on email change

edisonjpadilla:
Introduced changes: 
- Get the user by `uuid` instead of by email when the email is updated to prevent an eventual consistency error
- Update the `bridge_user` respecting how the shared workspace works: 
  - When the user is a host: update the `username`, the `email`, and the `brigde_user` of itself and all the guests
  - When the user is a guest: update the `username` and the `email` only 
  - When there is no shared workspace, update the three fields (just for that user)

### DIFF
--- a/src/modules/user/user.usecase.spec.ts
+++ b/src/modules/user/user.usecase.spec.ts
@@ -1,5 +1,7 @@
 import { Test, TestingModule } from '@nestjs/testing';
 import { createMock } from '@golevelup/ts-jest';
+import { AttemptChangeEmailModel } from './attempt-change-email.model';
+import { UserEmailAlreadyInUseException } from './exception/user-email-already-in-use.exception';
 
 import { UserUseCases } from './user.usecase';
 import { ShareUseCases } from '../share/share.usecase';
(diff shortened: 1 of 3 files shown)


src/modules/user/user.usecase.spec.ts
```diff
@@ -217,6 +248,244 @@ describe('User use cases', () => {
       });
     });
   });
+
+  describe('changeUserEmailById', () => {
+    it('should change the user email successfully', async () => {
```
apsantiso, line 253: Using "Should" is pretty valid, however, let's follow the same naming convention that is being used across the other tests. We should follow the same pattern even if it sounds too repetitive.

**"When X, then X"**   Check the tests made for Resseting a user.

src/modules/user/user.usecase.spec.ts
```diff
@@ -217,6 +248,244 @@ describe('User use cases', () => {
...
+      jest.spyOn(bridgeService, 'updateUserEmail').mockResolvedValue(undefined);
+      jest
+        .spyOn(sharedWorkspaceRepository, 'updateGuestEmail')
+        .mockResolvedValue(undefined);
+      jest.spyOn(userRepository, 'updateBy').mockResolvedValue(undefined);
```
apsantiso, line 258: Mocking things like "`updateBy`" and "`updateById`" just to return undefined will make test files grow larger than needed. Try to only Spy on objects that you really need to return `undefined` to make tests easier to read.  

These tests run fine without spying on every "update xxx" call. Only spy on functions that you either require to check how it has been called or are essential. I know `findByUsername` and other `get functions` need to be mocked to undefined because by default it returns` {}` which is considered as an existent object so don't worry about them.

apsantiso (approved):

src/modules/user/user.usecase.spec.ts
sg-gs, line 1: I am approving this PR, but except for the describe header, please update the legends of the tests so they explain what should happen and not how. For instance: 

'When the user is a guest on a shared workspace, Then it should not call bridgeService.updateUserEmail'

Replace it for something like: 

'When the user is a guest on a shared workspace, then the network email is not updated'

Why? 
If the method you are referring to in the legend, changes its name in the future, neither the transpiler nor the execution will tell you that, so the legend would be misleading. Talking about behavior allows anyone to understand why the test fails and what is the test expecting: the network email should not be updated. 

If we write legends this way, there will be less things to maintain and keep track of, thus, reducing complexity.

sg-gs (approved):

sg-gs:
Please @edisonjpadilla, update the description to tell _briefly, but precisely,_ when / why the bridge_user update is "corresponding". 

Same with the PR header: handle-workspace-users-on-email-change will be more descriptive for future searches. Some title like that would be beneficial for us in case of future searches.

PR #264, opened 2024-01-24
[PB-845]: feat/Calculate-sharing-folder-size

edisonjpadilla:
This PR is to obtain the size of a public folder.

Currently filters `sharing` for the database by `ID`, `type`, and `itemType`; If it almost does not find a result, it throws a `SHARING_NOT_FOUND` error, if it exists, it continues with the request and returns the `size` of the folder.

### DIFF
--- a/src/common/uuid.dto.spec.ts
+++ b/src/common/uuid.dto.spec.ts
@@ -1,0 +1,33 @@
+import { validate } from 'class-validator';
+import { UuidDto } from './uuid.dto';
+import { newUser } from '../../test/fixtures';
+
+describe('UuidDto Validation', () => {
+  const user = newUser();
+
+  it('When a valid UUID is passed, then pass', async () => {
+    const dto = new UuidDto();
+    dto.id = user.uuid;
+
+    const errors = await validate(dto);
+    expect(errors.length).toBe(0);
+  });
+
+  it('When an invalid UUID is passed, then fail', async () => {
+    const dto = new UuidDto();
+    dto.id = 'invalid_uuid_string';
+
+    const errors = await validate(dto);
+    expect(errors.length).toBeGreaterThan(0);
+    expect(errors[0].constraints).toHaveProperty('isUuid');
+  });
+
+  it('When an empty string is passed, then fail', async () => {
+    const dto = new UuidDto();
+    dto.id = 'invalid-uuid';
+    const errors = await validate(dto);
+
+    expect(errors.length).toBeGreaterThan(0);
+    expect(errors[0].constraints).toHaveProperty('isUuid');
+  });
+});
--- a/src/modules/sharing/sharing.controller.spec.ts
+++ b/src/modules/sharing/sharing.controller.spec.ts
@@ -13,7 +13,7 @@ describe('SharingController', () => {
   beforeEach(async () => {
     sharingService = createMock<SharingService>();
     controller = new SharingController(sharingService);
-    sharing = newSharing()
+    sharing = newSharing({})
   });
 
   describe('get public sharing folder size', () => {
--- a/src/modules/sharing/sharing.controller.spec.ts
+++ b/src/modules/sharing/sharing.controller.spec.ts
@@ -1,0 +1,37 @@
+import { SharingController } from './sharing.controller';
+import { SharingService } from './sharing.service';
+import { UuidDto } from '../../common/uuid.dto';
+import { createMock } from '@golevelup/ts-jest';
+import { Sharing } from './sharing.domain'
+import { newSharing } from '../../../test/fixtures';
+
+describe('SharingController', () => {
+  let controller: SharingController;
+  let sharingService: SharingService;
+  let sharing: Sharing;
+
+  beforeEach(async () => {
+    sharingService = createMock<SharingService>();
+    controller = new SharingController(sharingService);
+    sharing = newSharing()
+  });
+
+  describe('get public sharing folder size', () => {
+    it('When request the get sharing size method, then it works', async () => {
+      const expectedResult = 100;
+
+      jest
+        .spyOn(sharingService, 'getPublicSharingFolderSize')
+        .mockImplementation(async () => expectedResult);
+
+      const result = await controller.getPublicSharingFolderSize({
+        id: sharing.id,
+      } as UuidDto);
+
+      expect(result).toStrictEqual({ size: expectedResult });
+      expect(sharingService.getPublicSharingFolderSize).toHaveBeenCalledWith(
+        sharing.id,
+      );
+    });
+  });
+});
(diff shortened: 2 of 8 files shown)


src/modules/sharing/sharing.service.ts
```diff
@@ -2057,4 +2058,14 @@ export class SharingService {
 
     return sharedItem;
   }
+
+  async getSharingFolderSize(id: SharingAttributes['id']): Promise<number> {
+    const sharing = await this.sharingRepository.findOneSharing({ id });
```
apsantiso, line 2063: Let's check if it is public. For public sharings, we basically add  sharedWith with the value  `00000000-0000-0000-0000-000000000000`, or so.

You can check in the endpoint GET sharing/meta how it works. So, check by id and sharedWith: `00000000-0000-0000-0000-000000000000`.  If there is no sharing with those conditions, then that sharing is not public, so you can return throw new SharingNotFoundException(); as you are doing right now.

src/modules/sharing/sharing.controller.ts
```diff
@@ -1026,4 +1029,12 @@ export class SharingController {
 
     return { message: 'User removed from shared folder' };
   }
+
+  @UseFilters(new HttpExceptionFilter())
+  @UseGuards(ThrottlerGuard)
+  @Public()
+  @Get(':id/size')
```
apsantiso, line 1036: Just to be a little more clear that this is only for public shared folders, could you change it to something like
` @Get('/public/:sharingId/folder/size')` ? You can use any naming you consider better anyways, we just need to make clear that this is not going to return file size or private sharings size.

src/common/uuid.dto.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { validate } from 'class-validator';
+import { UuidDto } from './uuid.dto';
+
+describe('UuidDtoDto Validation', () => {
+  it('should pass validation with a valid UUID', async () => {
```
apsantiso, line 5: Remember the "When xxxx, then xxx" structure. I know this is just for a dto, but just to be consistent 👍🏻

apsantiso: When a valid UUID is passed, then pass.

When an invalid UUID is passed, then fail.

When an empty string is passed, then fail.

apsantiso (changes requested):

src/modules/sharing/sharing.service.ts
apsantiso: @sg-gs I made this review assuming we want this endpoint only for public sharings as it is public. If you want an endpoint that returns the size ignoring the sharing type, then please @edisonjpadilla, ignore the comment related the `sharedWith` field.

src/modules/sharing/sharing.controller.ts
edisonjpadilla: it's for shared files, could be public or private

src/common/uuid.dto.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { validate } from 'class-validator';
+import { UuidDto } from './uuid.dto';
+
+describe('UuidDtoDto Validation', () => {
```
sg-gs, line 4: Typo here

src/modules/sharing/sharing.controller.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  let sharingService: SharingService;
+
+  beforeEach(async () => {
+    sharingService = createMock<SharingService>();
+    controller = new SharingController(sharingService);
+  });
+
+  describe('get sharing size', () => {
+    it('When request the get sharing size method, then it works', async () => {
+      const sharingId = 'valid-sharing-id';
```
sg-gs, line 18: let's use `chance` library to generate a valid sharing id (a uuid) so we validate that the format control is working properly. This should be already done if you export the fixture for sharing and you use it for generating a new one

src/modules/sharing/sharing.service.spec.ts
```diff
@@ -491,4 +493,71 @@ describe('Sharing Use Cases', () => {
...
+    it('When user tries to get sharing folder size with a public folder, then it works', async () => {
+      const sharing = newSharing({
+        owner,
+        item: folder,
+        sharedWith: otherUser,
+        sharingType: SharingType.Public,
+      });
+
+      sharingRepository.findOneSharing.mockResolvedValue(sharing);
+      folderUseCases.getFolderSizeByUuid.mockResolvedValue(200);
```
sg-gs, line 511: Let's extract 200 in a variable so if you change it, you do not have to change it in two different points. Also, I though at first it was a HTTP status code, so the variable naming would help a lot there

src/modules/sharing/sharing.controller.ts
sg-gs: This should not be the case. A private sharing should be protected under auth checks like a logged user, that has access (because is the `owner` or the `sharedWith` user) to the metadata explicitly.

src/modules/sharing/sharing.service.ts
```diff
@@ -2057,4 +2059,21 @@ export class SharingService {
...
+
+    if (!sharing) {
+      throw new SharingNotFoundException();
+    }
+
+    if (sharing.itemType !== 'folder') {
+      throw new InvalidSharingException('Item is not a folder');
+    }
+
+    return await this.folderUsecases.getFolderSizeByUuid(sharing.itemId);
```
sg-gs, line 2077: `return` directly, `await` is not needed in this line

sg-gs (changes requested):

sg-gs:
Where is the description of this PR @edisonjpadilla?

src/modules/sharing/sharing.controller.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+  let sharingService: SharingService;
+
+  beforeEach(async () => {
+    sharingService = createMock<SharingService>();
+    controller = new SharingController(sharingService);
+  });
+
+  describe('get public sharing folder size', () => {
+    it('When request the get sharing size method, then it works', async () => {
+      const sharingId = new Chance().guid();
```
sg-gs, line 18: Hey @edisonjpadilla, let's use the fixtures, therefore you have the centralization in that generator (`newSharing`). If we change the format of the id or whatever, you will need to change all the references in the code.

src/common/uuid.dto.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { validate } from 'class-validator';
+import { UuidDto } from './uuid.dto';
+
+describe('UuidDto Validation', () => {
+  it('When a valid UUID is passed, then pass', async () => {
+    const dto = new UuidDto();
+    dto.id = '550e8400-e29b-41d4-a716-446655440000';
```
sg-gs, line 7: Use the fixture here also, please

sg-gs (approved):

sg-gs (approved):

PR #265, opened 2024-01-29
[_]: feat/check if the user exists or has sub

masterprog-cmd:
A new endpoint has been added to check if the user exists or not and, if the user exists, to check if the user has a subscription.

### DIFF
--- a/src/modules/user/user.controller.ts
+++ b/src/modules/user/user.controller.ts
@@ -189,7 +189,7 @@ export class UserController {
     try {
       const user = await this.userUseCases.getUserByUsername(email);
       if (!user) {
-        throw new NotFoundException('PRE_CREATED_USER_NOT_FOUND');
+        throw new NotFoundException();
       }
 
       const userHasSubscriptions =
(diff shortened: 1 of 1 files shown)


sg-gs:
Hey @masterprog-cmd, 
This endpoint can not be public. Let's make it private as a GET and we can offer it as an endpoint callable by the _user_. 

Then, we can mock the user token from anywhere signing it with the same secret (for example, from any of our backends, like the website one). 

Otherwise, we would be exposing delicate info from our customers.

sg-gs (changes requested):
.

src/modules/user/user.controller.ts
```diff
@@ -161,6 +161,66 @@ export class UserController {
...
+    type: String,
+  })
+  async getUserByEmail(
+    @Param('email') email: User['email'],
+    @Res({ passthrough: true }) res: Response,
+  ) {
+    try {
+      const user = await this.userUseCases.getUserByUsername(email);
+      if (!user) {
+        throw new NotFoundException('PRE_CREATED_USER_NOT_FOUND');
```
sg-gs, line 192: This is wrong, precreated users are for invited users to a private sharing, this is not related. Just return a NotFound without any message by parameter

sg-gs (changes requested):

sg-gs (approved):

PR #266, opened 2024-02-02
[PB-1433]: feat/add-feature-limit-guard

apsantiso:
This PR is a POC for the creation of a guard that let us to check feature limits without introducing almost zero changes on services unless it is necessary. 

Let's consider that there are 3 types of limits (from a technical view):
1. Counter Limit that is not bound to any item. You can check it just with the user.uuid.
2. Counter Limit bound to X item. You need to get the itemId or whatever you need from the request, it is not enough with just the user.uuid
3. Boolean Limit: these can be checked without any extra logic.

## Added
1. **ApplyLimit**  decorator: This decorator is used to configure the limit to be applied. There are limits that need data from the body, query, params, header to be able to check them. This decorator is on charge of defining where should we get the data from. It does not do any check by itself, data is checked and extracted in the guard.
4. **FeatureLimit** guard: It is on charge of extracting the data from the context-request and validate everything is OK before proceeding to send data to the **limit checker service**.
5. **FeatureLimitUsecases**:  contains the functions that are used to check every limit. This service should have injected the different repositories, etc that needs for its functionality, so we can encapsulate bussiness logic in it.
6. **Temporal migration service**: Added a service that connects with payments API to get users current plan.
7. **Paid Plans table**: Table created to Map paid plans (stripe) with feature limit tiers.
8. Migrations for basic tiers and their limits according to discussed limits. [here](https://inxt.atlassian.net/wiki/spaces/PROD/pages/153321481/Individual+accounts)
9. Added free tier to users on sign up.

### What if I want to add a new limit?
New Limits can be added. You only need to consider these two points:
1. Counter Limit = Should have a function to be checked against assigned in LimitCheckService, otherwise, the guard is going to throw an error.
2. Boolean Limit = It is not necessary to add any function, the type itself just checks if value is 1 or 0.

### What if an endpoint needs depends on more than 1 limit?
The guard decorator allows you to send a string of limitLabels, so you are able to check multiple limits at one. However, the dataSources should be consistent. 

### Can I call the service to check a limit without calling the guard?
Yes, the good thing of encapsulating the logic in this service is that you can call the service to check anything you need. As long as you pass the correct arguments.

For example, you could call  `FeatureLimitUsecases.checkMaxSharedItemsLimit({data... user... limit...})` from your controller to check whatever you need without any decorator or guard.

### ApplyLimit

```javascript

  @Post('/invites/send')
  @ApplyLimit({
    limitLabels: [LimitLabels.MaxSharedItemInvites, LimitLabels.MaxSharedItems],
    dataSources: [
      { sourceKey: 'body', fieldName: 'itemId' },
      { sourceKey: 'body', fieldName: 'itemType' },
    ],
  })
  @UseGuards(FeatureLimit)
  createInvite(
    @UserDecorator() user: User,
    @Body() createInviteDto: CreateInviteDto,
  ) {
    return this.sharingService.createInvite(user, createInviteDto);
  }
```

### DIFF
--- a/migrations/20240128230553-create-limits-table.js
+++ b/migrations/20240128230553-create-limits-table.js
@@ -20,7 +20,7 @@ module.exports = {
         allowNull: false,
       },
       value: {
-        type: Sequelize.INTEGER,
+        type: Sequelize.STRING,
         allowNull: false,
       },
       created_at: {
--- a/migrations/20240128230553-create-limits-table.js
+++ b/migrations/20240128230553-create-limits-table.js
@@ -1,0 +1,42 @@
+'use strict';
+
+const tableName = 'limits';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.createTable(tableName, {
+      id: {
+        type: Sequelize.UUID,
+        allowNull: false,
+        primaryKey: true,
+      },
+      label: {
+        type: Sequelize.STRING,
+        allowNull: false,
+      },
+      type: {
+        type: Sequelize.ENUM('counter', 'boolean'),
+        allowNull: false,
+      },
+      value: {
+        type: Sequelize.INTEGER,
+        allowNull: false,
+      },
+      created_at: {
+        type: Sequelize.DATE,
+        allowNull: false,
+        defaultValue: Sequelize.NOW,
+      },
+      updated_at: {
+        type: Sequelize.DATE,
+        allowNull: false,
+        defaultValue: Sequelize.NOW,
+      },
+    });
+  },
+
+  async down(queryInterface) {
+    await queryInterface.dropTable(tableName);
+  },
+};
--- a/src/modules/feature-limit/feature-limit-migration.service.ts
+++ b/src/modules/feature-limit/feature-limit-migration.service.ts
@@ -5,7 +5,7 @@ import { AxiosError } from 'axios';
 import { User } from '../user/user.domain';
 import { PaymentsService } from 'src/externals/payments/payments.service';
 
-const FREE_TIER_ID = 'free';
+const FREE_TIER_ID = 'free_000000';
 
 @Injectable()
 export class FeatureLimitsMigrationService {
@@ -119,7 +119,7 @@ export class FeatureLimitsMigrationService {
 
     if (!this.planIdTierIdMap.get(FREE_TIER_ID)) {
       Logger.error(
-        `[FEATURE_LIMIT_MIGRATION/NO_FREE]: No free tier mapped found, please add a tier for free users`,
+        `[FEATURE_LIMIT_MIGRATION/NO_FREE]: No free tier mapped found, please add a tier for free users to your DB`,
       );
       throw Error(
         'There is no free tier mapped, please add a tier for free users',
--- a/src/modules/feature-limit/feature-limit-migration.service.ts
+++ b/src/modules/feature-limit/feature-limit-migration.service.ts
@@ -1,11 +1,9 @@
 import { Injectable, Logger } from '@nestjs/common';
 import { SequelizeUserRepository } from '../user/user.repository';
 import { SequelizeFeatureLimitsRepository } from './feature-limit.repository';
-import { HttpClient } from 'src/externals/http/http.service';
-import { Sign } from 'src/middlewares/passport';
-import { ConfigService } from '@nestjs/config';
 import { AxiosError } from 'axios';
 import { User } from '../user/user.domain';
+import { PaymentsService } from 'src/externals/payments/payments.service';
 
 const FREE_TIER_ID = 'free';
 
@@ -20,8 +18,7 @@ export class FeatureLimitsMigrationService {
   constructor(
     private userRepository: SequelizeUserRepository,
     private tiersRepository: SequelizeFeatureLimitsRepository,
-    private httpClient: HttpClient,
-    private readonly configService: ConfigService,
+    private paymentsService: PaymentsService,
   ) {}
 
   async asignTiersToUsers() {
@@ -84,23 +81,11 @@ export class FeatureLimitsMigrationService {
   }
 
   private async getUserSubscription(user: User, retries = 0) {
-    const jwt = Sign(
-      { payload: { uuid: user.uuid } },
-      this.configService.get('secrets.jwt'),
-    );
-
-    const params = {
-      headers: {
-        'Content-Type': 'application/json',
-        Authorization: `Bearer ${jwt}`,
-      },
-    };
     try {
-      const res = await this.httpClient.get(
-        `${this.configService.get('apis.payments.url')}/subscriptions`,
-        params,
+      const subscription = await this.paymentsService.getCurrentSubscription(
+        user.uuid,
       );
-      return res.data;
+      return subscription;
     } catch (error) {
       if (error instanceof AxiosError) {
         if (error.response && error.response.status === 404) {
--- a/src/modules/feature-limit/feature-limit-migration.service.ts
+++ b/src/modules/feature-limit/feature-limit-migration.service.ts
@@ -1,0 +1,148 @@
+import { Injectable, Logger } from '@nestjs/common';
+import { SequelizeUserRepository } from '../user/user.repository';
+import { SequelizeFeatureLimitsRepository } from './feature-limit.repository';
+import { HttpClient } from 'src/externals/http/http.service';
+import { Sign } from 'src/middlewares/passport';
+import { ConfigService } from '@nestjs/config';
+import { AxiosError } from 'axios';
+import { User } from '../user/user.domain';
+
+const FREE_TIER_ID = 'free';
+
+@Injectable()
+export class FeatureLimitsMigrationService {
+  private readonly delayBetweenUsers = 10;
+  private readonly maxRetries = 3;
+  private readonly paymentsAPIThrottlingDelay = 60000;
+
+  private planIdTierIdMap: Map<string, string>; // PlanId/tierId mapping
+
+  constructor(
+    private userRepository: SequelizeUserRepository,
+    private tiersRepository: SequelizeFeatureLimitsRepository,
+    private httpClient: HttpClient,
+    private readonly configService: ConfigService,
+  ) {}
+
+  async asignTiersToUsers() {
+    const limit = 20;
+    let offset = 0;
+    let processed = 0;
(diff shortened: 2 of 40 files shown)


src/modules/feature-limit/limit-check.service.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  async isMaxInviteesPerItemsLimitExceeded({
+    limit,
+    data,
+  }: {
+    limit: Limit;
+    data: MaxInviteesPerItemAttribute;
+  }) {
+    const { itemId, itemType } = data;
```
apsantiso, line 70: We could add a checker for data here too. If something is missing, just log and throw with a meaningful message.

src/modules/sharing/sharing.module.ts
```diff
@@ -44,6 +46,6 @@ import { AppSumoModule } from '../app-sumo/app-sumo.module';
     SequelizeUserReferralsRepository,
     PaymentsService,
   ],
-  exports: [SharingService, SequelizeSharingRepository],
+  exports: [SharingService, SequelizeSharingRepository, SequelizeModule],
```
apsantiso, line 49: Exporting SequelizeModule is recommended by Nestjs in order to prevent failing due to model dependencies if you use repositories in other modules. 

https://docs.nestjs.com/techniques/database#models

If you want to use the repository outside of the module which imports SequelizeModule.forFeature, you'll need to re-export the providers generated by it. You can do this by exporting the whole module, like this:

src/modules/feature-limit/limit-check.service.ts
apsantiso: I'm also aware that we need to include the owner here. So we should add 1 extra to the count.

migrations/20240128230553-create-limits-table.js
```diff
@@ -0,0 +1,42 @@
...
+      label: {
+        type: Sequelize.STRING,
+        allowNull: false,
+      },
+      type: {
+        type: Sequelize.ENUM('counter', 'boolean'),
+        allowNull: false,
+      },
+      value: {
+        type: Sequelize.INTEGER,
```
sg-gs, line 23: This value is more flexible as a string. I got it, we can represent the boolean as 0,1, but in the hypothetic - and highly probable - case we need to set a string, this field would be useless. Leaving it as a string and then parsing the value depending on the type is more flexible in the long term.

src/modules/feature-limit/feature-limit.module.ts
```diff
@@ -0,0 +1,28 @@
...
+      TierModel,
+      Limitmodel,
+      TierLimitsModel,
+      TierLimitsModel,
+    ]),
+    forwardRef(() => SharingModule),
+  ],
+  providers: [
+    SequelizeFeatureLimitsRepository,
+    LimitCheckService,
```
sg-gs, line 23: Why not just `FeatureLimitUsecases`?

src/modules/feature-limit/feature-limit.repository.ts
```diff
@@ -0,0 +1,44 @@
...
+    @InjectModel(TierModel)
+    private tierModel: typeof TierModel,
+    @InjectModel(Limitmodel)
+    private limitModel: typeof Limitmodel,
+    @InjectModel(TierLimitsModel)
+    private tierLimitmodel: typeof TierLimitsModel,
+  ) {}
+
+  async findById(id: string) {
+    const tier = await this.tierModel.findByPk(id);
```
sg-gs, line 20: `return` directly, `async` and `await` are not needed here

src/modules/feature-limit/feature-limits.guard.ts
```diff
@@ -0,0 +1,84 @@
...
+    const request = context.switchToHttp().getRequest();
+    const user = request.user;
+    const metadata = this.reflector.get<ApplyLimitMetadata>(
+      FEATURE_LIMIT_KEY,
+      handler,
+    );
+
+    if (!metadata) {
+      new Logger().error(
+        `Missing metada for feature limit guard! url: ${request.url} handler: ${handler.name}`,
```
sg-gs, line 35: Typo here, `metadata` instead of `metada`

src/modules/feature-limit/limit.domain.ts
sg-gs, line 1: Hey, we want to encapsulate the business logic on the entities, ideally. That means we should not expose helpers to determine if the limit has been surpassed or not. Rather than that, the limit should have a method that requires whatever is the input, and inside the class, we determine those things, returning true or false depending on if the limit has been surpassed or not. 

If this is not done, future changes to the limit or how it is applied could affect non-directly related code (like middlewares or other use cases)

src/modules/feature-limit/limits.enum.ts
```diff
@@ -0,0 +1,9 @@
+export enum LimitLabels {
+  MaxSharedItems = 'max-shared-items',
+  MaxSharedItemInvites = 'max-shared-invites',
+}
+
+export enum LimitTypes {
+  Boolean = 'boolean',
+  counter = 'counter',
```
sg-gs, line 8: counter should be Counter as key, following the TypeScript official conventions, like you have done with Boolean, which has the first letter in uppercase.

src/modules/feature-limit/models/limit.model.ts
```diff
@@ -0,0 +1,50 @@
...
+export class Limitmodel extends Model implements LimitAttributes {
+  @PrimaryKey
+  @Column(DataType.UUIDV4)
+  id: string;
+
+  @AllowNull(false)
+  @Column(DataType.STRING)
+  label: LimitLabels;
+
+  @AllowNull(false)
```
sg-gs, line 29: I think a `@NotNull` decorator exists for this

src/modules/sharing/sharing.repository.ts
```diff
@@ -550,6 +550,16 @@ export class SequelizeSharingRepository implements SharingRepository {
     return SharingInvite.build(raw);
   }
 
+  async getSharedItemsNumberByUser(userUuid: string): Promise<number> {
+    const sharingsNumber = await this.sharings.count({
```
sg-gs, line 554: `sharingsCount` could be a better naming

src/modules/feature-limit/limit-check.service.ts
```diff
@@ -0,0 +1,101 @@
...
+          itemId,
+          itemType,
+        }),
+      ]);
+
+    const count = sharingsCountForThisItem + invitesCountForThisItem;
+
+    const limitExceeded = count >= limit.value;
+    if (limitExceeded) {
+      throw new BadRequestException(
```
sg-gs, line 88: a 402 HTTP status code would be more descriptive, also allowing the frontend to identify better the limit error

sg-gs (commented):
The POC is fine, the concept seems scalable, and is using the Nestjs Guards and Decorators at its finest. Just keep an eye on not coupling them too much or revealing the limits details of implementation to other use cases, as that could get out of control fast

src/modules/feature-limit/models/limit.model.ts
apsantiso: Yes, but the interpreter complains and the app does not run. The message asks you to use @AllowNull(false) instead of it. Pretty strange imo.

src/modules/feature-limit/feature-limit-migration.service.ts
```diff
@@ -0,0 +1,132 @@
...
+    private paymentsService: PaymentsService,
+  ) {}
+
+  async asignTiersToUsers() {
+    const limit = 20;
+    let offset = 0;
+    let processed = 0;
+    await this.loadTiers();
+
+    while (true) {
```
sg-gs, line 29: Hey there is a way to do this without a true and a break. Let's use a bit of math here:

```
let resultsCount = 0;

while (resultsCount % limit === 0) {
 ....
 resultsCount = users.length;
}
```

This is dynamic with the limit and handles cases where `users.length` could be different than 0, for instance: the last page of the pagination.

src/modules/feature-limit/feature-limit-migration.service.ts
```diff
@@ -0,0 +1,132 @@
...
+      for (const user of users) {
+        await this.assignTier(user);
+        await this.delay(this.delayBetweenUsers); // Delay between requests to prevent Stripe Throttling
+      }
+
+      processed += users.length;
+      Logger.log(
+        `[FEATURE_LIMIT_MIGRATION]: Processed : ${processed}, current offset: ${offset} `,
+      );
+      offset += limit;
```
sg-gs, line 49: If you are paginating by `tierId: null` and you are assigning tier to the current chunk of users, the offset is not needed as the next page with the WHERE clause will exclude those newly assigned tier users from the result.

In any case, the offset would be useful for failed cases in the case we want to advance no matter the situation. (`offset += failedUsers.length`)

src/modules/feature-limit/feature-limit-migration.service.ts
```diff
@@ -0,0 +1,132 @@
...
+    try {
+      const subscription = await this.paymentsService.getCurrentSubscription(
+        user.uuid,
+      );
+      return subscription;
+    } catch (error) {
+      if (error instanceof AxiosError) {
+        if (error.response && error.response.status === 404) {
+          return;
+        } else if (error.response && error.response.status === 429) {
```
sg-gs, line 92: The first if returns, therefore, you can just replace the else if by an if.

sg-gs (approved):

sg-gs:
Migrations run. 

```
== 20240128230532-create-tiers-table: migrating =======
== 20240128230532-create-tiers-table: migrated (0.185s)

== 20240128230553-create-limits-table: migrating =======
== 20240128230553-create-limits-table: migrated (0.279s)

== 20240128230608-create-tiers-limits-table: migrating =======
== 20240128230608-create-tiers-limits-table: migrated (0.195s)

== 20240128230626-add-tier-id-to-users: migrating =======
== 20240128230626-add-tier-id-to-users: migrated (0.248s)

== 20240220163114-create-paid-plans-tiers: migrating =======
== 20240220163114-create-paid-plans-tiers: migrated (0.253s)

== 20240220203734-seed-tiers-and-limits: migrating =======
== 20240220203734-seed-tiers-and-limits: migrated (0.296s)

== 20240220215830-seed-free-plan-to-paid-plans-table: migrating =======
== 20240220215830-seed-free-plan-to-paid-plans-table: migrated (0.190s)

✨  Done in 5.13s
```

sg-gs:
New environment variable `PAYMENTS_API_URL` set.

PR #272, opened 2024-02-23
[PB-1706]: Retrieve subitems by folder UUID

larry-internxt:
- Added missing parentUuid property to folders
- Added getFolderContentFolders function to get the subfolders from a folder
- Added getFolderContentFiles function to get the subfiles from a folder
- Added unit tests

### DIFF
--- a/src/modules/folder/folder.controller.spec.ts
+++ b/src/modules/folder/folder.controller.spec.ts
@@ -135,5 +135,133 @@ describe('FolderController', () => {
 
       expect(result).toEqual({ folders: mappedSubfolders });
     });
+
+    it('When get folder subfiles are requested by invalid params, then it should throw an error', async () => {
+      try {
+        await folderController.getFolderContentFiles(
+          userMocked,
+          'invalidUUID',
+          50,
+          0,
+          'id',
+          'ASC',
+        );
+        expect(true).toBeFalsy; // getFolderContentFiles should throw an error
+      } catch (err) {
+        expect((err as Error).message).toEqual('Invalid UUID provided');
+      }
+
+      try {
+        await folderController.getFolderContentFiles(
+          userMocked,
+          folder.uuid,
+          0,
+          0,
+          'id',
+          'ASC',
+        );
+        expect(true).toBeFalsy; // getFolderContentFiles should throw an error
+      } catch (err) {
+        expect((err as Error).message).toEqual(
+          'Limit should be between 1 and 50',
+        );
+      }
+
+      try {
+        await folderController.getFolderContentFiles(
+          userMocked,
+          folder.uuid,
+          51,
+          0,
+          'id',
+          'ASC',
+        );
+        expect(true).toBeFalsy; // getFolderContentFiles should throw an error
+      } catch (err) {
+        expect((err as Error).message).toEqual(
+          'Limit should be between 1 and 50',
+        );
+      }
+
+      try {
+        await folderController.getFolderContentFiles(
+          userMocked,
+          folder.uuid,
+          50,
+          -1,
+          'id',
+          'ASC',
+        );
+        expect(true).toBeFalsy; // getFolderContentFiles should throw an error
+      } catch (err) {
+        expect((err as Error).message).toEqual(
+          'Offset should be higher than 0',
+        );
+      }
+    });
+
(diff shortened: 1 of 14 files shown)


PixoDev (dismissed):

apsantiso (approved):

PixoDev (dismissed):

src/modules/folder/folder.controller.spec.ts
```diff
@@ -47,4 +81,187 @@ describe('FolderController', () => {
...
+      try {
+        await folderController.getFolderContentFiles(
+          userMocked,
+          folder.uuid,
+          0,
+          0,
+          'id',
+          'ASC',
+        );
+        expect(true).toBeFalsy; // getFolderContentFiles should throw an error
```
sg-gs, line 163: The comment is not adding something difficult to understand, the code is semantic enough to understand that "the test expects the function to throw an error with a specific message. 

By the way, the most maintainable solution to match errors is using `instanceof` as if the limit minimum or maximum numbers change, the message is prone to change, so this test would need to be updated. So, it would be better to find something like: 

`expect(err instanceof TheErrorIWantToMatch).isTruthy();`

src/modules/folder/folder.controller.spec.ts
```diff
@@ -47,4 +81,187 @@ describe('FolderController', () => {
...
+          userMocked,
+          folder.uuid,
+          51,
+          0,
+          'id',
+          'ASC',
+        );
+        expect(true).toBeFalsy; // getFolderContentFiles should throw an error
+      } catch (err) {
+        expect((err as Error).message).toEqual(
```
sg-gs, line 181: Same here

src/modules/folder/folder.controller.spec.ts
```diff
@@ -47,4 +81,187 @@ describe('FolderController', () => {
...
+          userMocked,
+          folder.uuid,
+          50,
+          -1,
+          'id',
+          'ASC',
+        );
+        expect(true).toBeFalsy; // getFolderContentFiles should throw an error
+      } catch (err) {
+        expect((err as Error).message).toEqual(
```
sg-gs, line 197: And here

src/modules/folder/folder.controller.spec.ts
```diff
@@ -47,4 +81,187 @@ describe('FolderController', () => {
...
+          userMocked,
+          'invalidUUID',
+          50,
+          0,
+          'id',
+          'ASC',
+        );
+        expect(true).toBeFalsy; // getFolderContentFolders should throw an error
+      } catch (err) {
+        expect((err as Error).message).toEqual('Invalid UUID provided');
```
sg-gs, line 215: And here

src/modules/folder/folder.controller.spec.ts
```diff
@@ -47,4 +81,187 @@ describe('FolderController', () => {
...
+          userMocked,
+          folder.uuid,
+          0,
+          0,
+          'id',
+          'ASC',
+        );
+        expect(true).toBeFalsy; // getFolderContentFolders should throw an error
+      } catch (err) {
+        expect((err as Error).message).toEqual(
```
sg-gs, line 229: Here too

src/modules/folder/folder.controller.spec.ts
```diff
@@ -47,4 +81,187 @@ describe('FolderController', () => {
...
+          userMocked,
+          folder.uuid,
+          51,
+          0,
+          'id',
+          'ASC',
+        );
+        expect(true).toBeFalsy; // getFolderContentFolders should throw an error
+      } catch (err) {
+        expect((err as Error).message).toEqual(
```
sg-gs, line 245: Here too

src/modules/folder/folder.controller.spec.ts
```diff
@@ -47,4 +81,187 @@ describe('FolderController', () => {
...
+          userMocked,
+          folder.uuid,
+          50,
+          -1,
+          'id',
+          'ASC',
+        );
+        expect(true).toBeFalsy; // getFolderContentFolders should throw an error
+      } catch (err) {
+        expect((err as Error).message).toEqual(
```
sg-gs, line 261: Here too

sg-gs (dismissed):
Fine for me, just adapt those error matchings on the test please @larry-internxt

sg-gs (approved):

PR #277, opened 2024-03-01
[PB-1740]: feat/add items to trash by uuid

apsantiso:
This PR allows to add files by uuid to trash.

Either id or uuid should be provided inside the items. If none is provided, then the input is invalid.

### DIFF
--- a/src/modules/file/file.usecase.ts
+++ b/src/modules/file/file.usecase.ts
@@ -248,8 +248,12 @@ export class FileUseCases {
   moveFilesToTrash(
     user: User,
     fileIds: FileAttributes['fileId'][],
-  ): Promise<void> {
-    return this.fileRepository.updateFilesStatusToTrashed(user, fileIds);
+    fileUuids: FileAttributes['uuid'][] = [],
+  ): Promise<[void, void]> {
+    return Promise.all([
+      this.fileRepository.updateFilesStatusToTrashed(user, fileIds),
+      this.fileRepository.updateFilesStatusToTrashedByUuid(user, fileUuids),
+    ]);
   }
 
   async getEncryptionKeyFileFromShare(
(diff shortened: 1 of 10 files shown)


src/modules/file/file.usecase.ts
```diff
@@ -246,8 +246,12 @@ export class FileUseCases {
   moveFilesToTrash(
     user: User,
     fileIds: FileAttributes['fileId'][],
-  ): Promise<void> {
-    return this.fileRepository.updateFilesStatusToTrashed(user, fileIds);
+    fileUuids: FileAttributes['uuid'][] = [],
+  ): Promise<[void, void]> {
+    return Promise.all([
```
sg-gs, line 251: Hey, we do not want to move items that already trashed into the trash again, therefore, you need to add:
`where status = "EXISTS"`
So you can prevent concurrency issues if multiples devices are syncing.

src/modules/folder/folder.usecase.ts
```diff
@@ -294,12 +294,18 @@ export class FolderUseCases {
   async moveFoldersToTrash(
     user: User,
     folderIds: FolderAttributes['id'][],
+    folderUuids: FolderAttributes['uuid'][] = [],
   ): Promise<void> {
-    const [folders, driveRootFolder] = await Promise.all([
+    const [foldersById, driveRootFolder, foldersByUuid] = await Promise.all([
       this.getFoldersByIds(user, folderIds),
       this.getFolder(user.rootFolderId),
+      folderUuids.length > 0
```
sg-gs, line 302: Hey, we do not want to move items that already trashed into the trash again, therefore, you need to add:
`where deleted = "false" and removed = "false"`
So you can prevent concurrency issues if multiples devices are syncing.

sg-gs (changes requested):

src/modules/folder/folder.usecase.ts
apsantiso: Gotcha, added the conditions to the repository so trashed  folders are not fetch.

sg-gs (approved):

PR #278, opened 2024-03-04
[PB-1429]: Feat/max storage days limit

apsantiso:
This PR enforces the limit for the storage days in the trash.

What it includes:

1. Do not list expired trashed items in get trashed items endpoint.
2. Add a cron job to expire files every X defined time.

Some considerations:

- Cron jobs with multiple instances running: nestjs cron jobs are in memory, every instance is going to try to run the cron job. There are some ways to handle this, having a table in a shared db so the first instance that tries to run the job writes to the DB with the current timestamp (it sounds kind of a race condition I don't like it) or using redis. 

   > However, for our case, maybe it is too much trouble to handle this simple cron job that eventually is going to process a small quantity of files. It runs every few minutes, we are not going to have a lot of expired trashed files in such a small time frame.

- Folders use deleteAt and files use updatedAt. Should we index both? Should we update "updatedAt" for both when they are trashed? 

- Why is there a raw query inside the query for listing trashed/expired items? Actually, I tried to use the ORM at its max extend without writing raw SQL (they are too troublesome to maintain 😴  and people sometimes don't check them before changing models). This was my original query:

```javascript
    const files = await this.fileModel.findAll({
      attributes: ['status', 'plainName', 'updatedAt', 'fileId', 'uuid'],
      where: {
        status: 'TRASHED',
        updatedAt: {
          [Op.lt]: Sequelize.literal(
            `NOW() - "user->tier->limits"."value"::integer * INTERVAL '1 day'`,
          ),
        },
      },
      include: {
        model: UserModel,
        attributes: ['uuid', 'id', 'tierId'],
        as: 'user',
        include: [
          {
            model: TierModel,
            attributes: ['id', 'label'],
            include: [
              {
                model: Limitmodel,
                where: { label: LimitLabels.MaxTrashStorageDays },
              },
            ],
          },
        ],
      },
      limit,
    });
```

However, Sequelize seems to have a [known bug](https://github.com/sequelize/sequelize/issues/3623) when you try to nest models with many to many relations sometimes. Basically, the query works if I do not add the `limit` at the end to just get a X number of records OR if I do not ask for the `limitModel` while limiting the results. It creates a sort of subquery and introduces the limit = 1000 inside that subquery and it fails.

Of course, I do not want to get the entire database of expired trashed items, so that's the reason of the raw query being there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "@nestjs/jwt": "^10.2.0",
     "@nestjs/passport": "^10.0.2",
     "@nestjs/platform-express": "^10.2.8",
+    "@nestjs/schedule": "^4.0.1",
     "@nestjs/sequelize": "^10.0.0",
     "@nestjs/swagger": "^7.1.16",
     "@nestjs/throttler": "^5.0.1",
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -23,6 +23,7 @@ import { FuzzySearchModule } from './modules/fuzzy-search/fuzzy-search.module';
 import { SharingModule } from './modules/sharing/sharing.module';
 import { AppSumoModule } from './modules/app-sumo/app-sumo.module';
 import { PlanModule } from './modules/plan/plan.module';
+import { ScheduleModule } from '@nestjs/schedule';
 
 @Module({
   imports: [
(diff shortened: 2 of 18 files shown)


apsantiso:
If we need to do this, there are better approaches.

PR #283, opened 2024-03-05
[PB-1768]: Feat/Move file & folders by UUID

larry-internxt:
Adds files and folders move endpoints using UUIDs

- Folders: PATCH /api/folders/:folderUuid
The flow is based on: https://github.com/internxt/drive-server/blob/6c5ff34eaaf1e4e712ab43abe174f992b93f46de/src/app/services/folder.js#L519

- Files: PATCH /api/files/:fileUuid
The flow is based on: https://github.com/internxt/drive-server/blob/6c5ff34eaaf1e4e712ab43abe174f992b93f46de/src/app/services/files.js#L281


apsantiso:
From my side, I see no problems other than the two I mentioned! It is just to prevent ghost files or something like that.

src/modules/file/file.usecase.ts
```diff
@@ -321,6 +322,66 @@ export class FileUseCases {
...
+    user: User,
+    fileUuid: File['fileId'],
+    destinationUuid: File['folderUuid'],
+  ): Promise<File> {
+    const file = await this.fileRepository.findByUuid(fileUuid, user.id);
+    if (!file) {
+      throw new NotFoundException(`File ${fileUuid} not found`);
+    }
+
+    const destinationFolder = await this.folderUsecases.getFolderByUuidAndUser(
```
sg-gs, line 335: Ensure the destination `folder.removed = false`. We cannot allow to move files to folders that are removed permanently

src/modules/file/file.usecase.ts
```diff
@@ -321,6 +322,66 @@ export class FileUseCases {
...
+      destinationUuid,
+      user,
+    );
+    if (!destinationFolder) {
+      throw new NotFoundException(`Folder ${destinationUuid} not found`);
+    }
+
```
sg-gs, line 342: Ensure the origin `folder.removed = false`. We cannot allow to move files from folders that are removed permanently

src/modules/folder/folder.controller.ts
```diff
@@ -525,4 +526,21 @@ export class FolderController {
 
     return { size };
   }
+
+  @Put('move/:folderUuid/to/:destinationUuid')
```
sg-gs, line 530: Same here, this is a partial update where you are updating a folder's attribute by uuid. It's more semantic to just use PATCH /:uuid, the new parentUuid should be sent on the body, and if you can use the field name as is, the better. If in the future we modify more parameters, we must be handling the usecase routing here.

sg-gs (changes requested):

apsantiso:
> > From my side, I see no problems other than the two I mentioned! It is just to prevent ghost files or something like that.
> > Once they are done, please tell @sg-gs
> 
> hey, i can't see any mention!

Sorry, I left my review still pending and I guess it does not appear to you, the same two that sergio made about omit deleted files and folders.

src/modules/file/file.controller.ts
```diff
@@ -175,4 +175,17 @@ export class FileController {
       return f;
     });
   }
+
+  @Put('move/:fileUuid/to/:folderUuid')
```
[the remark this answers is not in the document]
sg-gs: At the very end, we are updating a resource field, so following REST, if we are moving or not, it does not matter, we are doing a partial update of a resource, thus, we focus on what is happening with the resource (GET, POST, PUT, PATCH, DELETE and so on). This is a partial update, thus, a PATCH. Using /move on the path is adding unnecesary overhead from my opinion, and distracting developers to think in no-convention path naming instead of in RESTful terms

sg-gs (approved):

PR #291, opened 2024-03-15
[_] feat: workspace role guard

apsantiso:


src/modules/workspaces/guards/workspaces.guard.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (accessContext === AccessContext.WORKSPACE) {
+      return this.checkUserWorkspaceRole(user.uuid, id, requiredRole);
+    } else if (accessContext === AccessContext.TEAM) {
+      return this.checkUserTeamRole(user.uuid, id, requiredRole);
+    }
+
+    return false;
+  }
+
+  private async checkUserWorkspaceRole(
```
sg-gs, line 56: `enforceWorkspaceRolePolicy` could be a better naming

src/modules/workspaces/guards/workspaces.guard.ts
```diff
@@ -0,0 +1,114 @@
...
+    const { workspace, workspaceUser } =
+      await this.workspaceUseCases.findUserInWorkspace(userUuid, workspaceId);
+
+    if (!workspace) {
+      throw new NotFoundException('Workspace not found');
+    }
+
+    if (
+      !workspaceUser ||
+      !(role === WorkspaceRole.OWNER && workspace.isUserOwner(userUuid))
```
sg-gs, line 70: If you can extract this in a self-explainatory-named variable, it could improve the readability of this condition

src/modules/workspaces/guards/workspaces.guard.ts
```diff
@@ -0,0 +1,114 @@
...
+      Logger.log(
+        `[WORKSPACES/GUARD]: user has no requiered access to workspace. id: ${workspaceId} userUuid: ${userUuid} `,
+      );
+      throw new ForbiddenException('You have no access to this workspace');
+    }
+
+    return !!workspaceUser;
+  }
+
+  private async checkUserTeamRole(
```
sg-gs, line 81: `enforceTeamRolePolicy` could be a better naming

sg-gs (changes requested):

src/modules/workspaces/workspaces.controller.ts
```diff
@@ -65,7 +76,9 @@ export class WorkspacesController {
     return this.workspaceUseCases.getWorkspaceTeams(user, workspaceId);
   }
 
-  @Get('/:workspaceId/teams/:teamId/members')
+  @Get('/teams/:teamId/members')
```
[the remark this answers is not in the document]
apsantiso: No, I just removed the workspace because the teams have unique IDs. However, let's add it again just to be able to enforce any type of check just in case.

src/modules/workspaces/guards/workspaces.guard.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    return false;
+  }
+
+  private async checkUserWorkspaceRole(
+    userUuid: string,
+    workspaceId: string,
+    role: WorkspaceRole,
+  ) {
+    const { workspace, workspaceUser } =
```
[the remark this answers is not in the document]
apsantiso: Either the workspace or the workspaceUser can be null. I handled that case in the usecase so a plain and not destructurable null is never returned.

src/modules/workspaces/guards/workspaces.guard.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+      const workspace = newWorkspace();
+
+      jest.spyOn(reflector, 'get').mockReturnValue({
+        requiredRole: WorkspaceRole.MEMBER,
+        accessContext: AccessContext.WORKSPACE,
+        idSource: 'params',
+      });
+      workspaceUseCases.findUserInWorkspace.mockResolvedValue({
+        workspace,
+        workspaceUser: {} as WorkspaceUser,
```
sg-gs, line 138: Why not setting `workspaceUser: workspaceMember`?

src/modules/workspaces/guards/workspaces.guard.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+        query: {},
+        body: {},
+      });
+
+      await expect(guard.canActivate(context)).rejects.toThrow(
+        NotFoundException,
+      );
+    });
+
+    it('Access Granted for Team Member', async () => {
```
sg-gs, line 196: When-then not used here

src/modules/workspaces/guards/workspaces.guard.ts
```diff
@@ -0,0 +1,130 @@
+// flexible-access.guard.ts
```
sg-gs, line 1: Remove this comment

sg-gs (approved):

PR #292, opened 2024-03-20
[PB-1551]: Feat/invite users to org

apsantiso:
This PR is to enable workspace invitations to either new users and already registered users.

Mnemonic should come encrypted from the frontend with the public key of the precreated user or the registered email. Same pattern as we are using for sharing.

Invites should be only accepted if there is enough same left in the workspace. Once an invitation is sent, then that space counts as assigned.

**As defined for the MVP,  I set the limit to 10 users per workspace.** 

New Environment variables for sengrid templates:

```
#Workspaces
WORKSPACES_USER_INVITATION_EMAIL_ID=d-de1ed6df4a9947129c0bf592c808b58d
WORKSPACES_GUEST_USER_INVITATION_EMAIL_ID=d-41b4608fc94a41bca65aab7ed6ccad15
```


src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -0,0 +1,343 @@
...
+        service.inviteUserToWorkspace(user, workspace.id, {
+          invitedUser: 'test@example.com',
+          spaceLimit: BigInt(1024),
+          encryptionKey: 'encryptionKey',
+          encryptionAlgorithm: 'RSA',
+        }),
+      ).rejects.toThrow(BadRequestException);
+    });
+
+    it('When workspace has no more slots left, then it should throw', async () => {
```
sg-gs, line 175: This is the same test as above, isn't?

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -0,0 +1,343 @@
...
+        .spyOn(workspaceRepository, 'getWorkspaceUsersCount')
+        .mockResolvedValue(5);
+      jest
+        .spyOn(workspaceRepository, 'getWorkspaceInvitationsCount')
+        .mockResolvedValue(4);
+
+      const isFull = await service.isWorkspaceFull(workspaceId);
+      expect(isFull).toBe(false);
+    });
+    it('When workspace has slots left, then workspace is full', async () => {
```
sg-gs, line 295: I think the legend here is wrong, it should be: 
_When workspace does not have slots, left..._
Instead of: 
_When workspace has slots left_

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -0,0 +1,343 @@
...
+
+  it('should be defined', () => {
+    expect(service).toBeDefined();
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('inviteUserToWorkspace', () => {
```
sg-gs, line 66: I think there is a test case missing here: 
_When the invited user is not found, then it should throw_

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -0,0 +1,343 @@
...
+        .mockResolvedValue(BigInt(200000));
+
+      const assignableSpace = await service.getAssignableSpaceInWorkspace(
+        workspace,
+        workspaceDefaultUser,
+      );
+      expect(assignableSpace).toBe(BigInt(300000));
+    });
+
+    it('When there is no space left, then it should return 0', async () => {
```
sg-gs, line 327: As this function does not have multiple ways, just once, which is an aritmetical operation, this test is not covering anything new, as far as we check that the operation is being done properly (the first test)

sg-gs (changes requested):

sg-gs:
To improve the coverage, try to test at least on the repository the cases where a method should return 0 if the number being expected is null @apsantiso

PR #293, opened 2024-04-04
[PB-1920]: fix/stop sharing items when they are trashed

apsantiso:
This PR intends to stop sharing files/folders when you trash them. 

**Why**?

With features limits being implemented, we should stop sharing items when they are trashed to provide a smooth and clear experience. If I trash an item, I expect it to stop counting as a sharing.

**What could be improved?**

This PR adds this proccess in a sync way, however, we could also add an event listener in a future to decouple this process and make it async.

### DIFF
--- a/src/modules/file/file.usecase.ts
+++ b/src/modules/file/file.usecase.ts
@@ -13,7 +13,6 @@ import { CryptoService } from '../../externals/crypto/crypto.service';
 import { BridgeService } from '../../externals/bridge/bridge.service';
 import { FolderAttributes } from '../folder/folder.attributes';
 import { Share } from '../share/share.domain';
-import { ShareUseCases } from '../share/share.usecase';
 import { User } from '../user/user.domain';
 import { UserAttributes } from '../user/user.attributes';
 import {
@@ -27,6 +26,8 @@ import { SequelizeFileRepository } from './file.repository';
 import { FolderUseCases } from '../folder/folder.usecase';
 import { ReplaceFileDto } from './dto/replace-file.dto';
 import { FileDto } from './dto/file.dto';
+import { SharingService } from '../sharing/sharing.service';
+import { SharingItemType } from '../sharing/sharing.domain';
 
 type SortParams = Array<[SortableFileAttributes, 'ASC' | 'DESC']>;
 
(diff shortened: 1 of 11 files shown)


src/modules/file/file.usecase.ts
```diff
@@ -245,14 +246,22 @@ export class FileUseCases {
     return files.map((file) => file.toJSON());
   }
 
-  moveFilesToTrash(
+  async moveFilesToTrash(
     user: User,
     fileIds: FileAttributes['fileId'][],
     fileUuids: FileAttributes['uuid'][] = [],
-  ): Promise<[void, void]> {
+  ): Promise<[void, void, void]> {
```
sg-gs, line 253: `await` the promise.all and change the return type to Promise<void> instead of <[void, void, void]>

src/modules/sharing/sharing.service.ts
```diff
@@ -1460,6 +1464,19 @@ export class SharingService {
     });
   }
 
+  async bulkRemoveSharings(
+    user: User,
+    itemIds: Sharing['itemId'][],
+    itemType: Sharing['itemType'],
+  ) {
+    await this.sharingRepository.bulkDeleteInvites(itemIds, itemType);
```
sg-gs, line 1472: How are you ensuring that `bulkDeleteInvites` is not allowing a user who is not the owner of an item to delete an invitation of that item?

sg-gs (changes requested):

src/modules/sharing/sharing.service.ts
apsantiso: Hey @sg-gs, this is used in an internal method, so the functions that call it are checking that for us, but I get what you mean.

Unlike `sharings`, there is not a direct way to accomplish it with a simple DELETE, as we do not save the owner. If that item is already being shared (think of this as already having at least 1 row in `sharings` table), we can just use findOne to fetch 1 row and check the ownership before calling this.

However, 1 item can have rows in `sharing_invitations` without any in `sharings`. So we can add the file/folder relation to the sharingInvite model and make a DELETE with a join that checks correctly if user the owner of the folder or file. Also, we could fetch again the file/folder and check the ownership before deleting sharings, but we are already doing this on `moveFilesToTrash` and `moveFoldersToTrash`

sg-gs (approved):

PR #294, opened 2024-04-09
[PB-1841] feat: added change users role in workspace team

apsantiso:
This PR enables changing roles of users inside a workspace (teams). 

There are only two possible roles to be assigned in the workspace (excluding owner, which is non assignable):
- Manager of a team: it is able to manage the team users, remove them for the team, etc.
- Member: has access to the team but it is not able to manage users or settings of it.

If a manager is removed from his role, them the owner of the workspace becomes the manager of that team. Teams should always have manager.

### DIFF
--- a/src/modules/workspaces/workspaces.controller.ts
+++ b/src/modules/workspaces/workspaces.controller.ts
@@ -23,6 +23,8 @@ import {
   WorkspaceRole,
 } from './guards/workspace-required-access.decorator';
 import { CreateWorkspaceInviteDto } from './dto/create-workspace-invite.dto';
+import { WorkspaceTeamAttributes } from './attributes/workspace-team.attributes';
+import { ChangeUserRoleDto } from './dto/change-user-role.dto';
 
 @ApiTags('Workspaces')
 @Controller('workspaces')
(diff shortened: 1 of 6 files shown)


src/modules/workspaces/workspaces.controller.ts
```diff
@@ -98,4 +100,19 @@ export class WorkspacesController {
   async getTeamMembers() {
     throw new NotImplementedException();
   }
+
+  @Patch('/:workspaceId/teams/:teamId/members/role')
```
sg-gs, line 104: Why not `/:workspaceId/teams/:teamId/members/:memberId`?

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -311,6 +319,162 @@ describe('WorkspacesUsecases', () => {
     });
   });
 
+  describe('changeUserRole', () => {
+    it('When team does not exist, then error is throw', async () => {
```
sg-gs, line 323: typo here: then an error is thrown

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -311,6 +319,162 @@ describe('WorkspacesUsecases', () => {
...
+
+      await expect(
+        service.changeUserRole('workspaceId', 'teamId', {
+          userId: 'userId',
+          role: WorkspaceRole.MEMBER,
+        }),
+      ).rejects.toThrow(NotFoundException);
+    });
+
+    it('When user is not part of team, then error is throw', async () => {
```
sg-gs, line 334: Same here

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -311,6 +319,162 @@ describe('WorkspacesUsecases', () => {
...
+
+      await expect(
+        service.changeUserRole('workspaceId', 'teamId', {
+          userId: 'userId',
+          role: WorkspaceRole.MEMBER,
+        }),
+      ).rejects.toThrow(BadRequestException);
+    });
+
+    it('When user is member is updated to manager, then team manager is updated correctly', async () => {
```
sg-gs, line 371: "When a member of the team is upgrade to manager, then it works" could be better

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -311,6 +319,162 @@ describe('WorkspacesUsecases', () => {
...
+        userId: 'userId',
+        role: WorkspaceRole.MANAGER,
+      });
+
+      expect(teamRepository.updateById).toHaveBeenCalledWith(team.id, {
+        managerId: member.uuid,
+      });
+    });
+
+    it('When user is manager is updated to member, then owner is assigned as manager', async () => {
```
sg-gs, line 397: "When a team manager's role is changed to member, then the owner is assigned as manager" could be better

sg-gs (approved):

sg-gs:
Please, correct the typos, add a description to the PR, a label and an assignee (yourself) @apsantiso. Good job 👍🏽

PR #295, opened 2024-04-11
[_] feat: setup workspace

apsantiso:
We need an endpoint to setup the workspace after it has been initialized (it is initialized through the gateway). 

Here we add he required information for the workspace such as name, address, description and also link the owner user to the default team. The workspaceUser for the owner is also configured, for our convenience, the mnemonic is encrypted with the owner's public key, like all users in the workspace.

For this to work with our guard, I also added a check to deny access to users if the workspace is yet to be set up.


src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -101,8 +101,44 @@ export class WorkspacesUsecases {
...
+      spaceLimit: BigInt(0),
+      driveUsage: BigInt(0),
+      backupsUsage: BigInt(0),
+      key: setupWorkspaceDto.encryptedMnemonic,
+      deactivated: false,
+      createdAt: new Date(),
+      updatedAt: new Date(),
+    });
+
+    await this.workspaceRepository.addUserToWorkspace(workspaceUser);
```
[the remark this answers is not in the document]
apsantiso: Sorry this has to be drafted! However you are right, I need to check if workspace.owner === user before this. However, I think i will need to add the owner as workspaceUser in the initialize endpoint so the guard  in controllers can protect this endpoint successfully.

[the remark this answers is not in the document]
apsantiso: Solved this by adding an extra check to the guard and fetching the workspace with the respective filters to make sure it is the owner of it. Let me know if you were talking about another thing I missed.

src/modules/workspaces/guards/workspaces.guard.ts
```diff
@@ -73,8 +81,17 @@ export class WorkspaceGuard implements CanActivate {
...
+
+    if (isWorkspaceNotSetup && workspace.isUserOwner(user)) {
+      return true;
+    }
 
-    if (isUserNotInWorkspace || isRequiredOwnerRoleAndUserIsNotOwner) {
+    if (
+      isUserNotInWorkspace ||
+      isRequiredOwnerRoleAndUserIsNotOwner ||
+      isWorkspaceNotSetup
```
sg-gs, line 93: This is a bit misleading. In reality, you are rejecting when the workspace is not set up and the user is not the owner.

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -293,6 +293,83 @@ describe('WorkspacesUsecases', () => {
...
+      jest
+        .spyOn(workspaceRepository, 'findOne')
+        .mockResolvedValueOnce(workspace);
+
+      await service.setupWorkspace(user, 'workspace-id', {
+        name: 'Test Workspace',
+        encryptedMnemonic: 'encryptedMnemonic',
+      });
+
+      expect(workspaceRepository.addUserToWorkspace).toHaveBeenCalled();
```
sg-gs, line 322: Ensuring the function has been called with the owner's data would make this test stronger

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -293,6 +293,83 @@ describe('WorkspacesUsecases', () => {
...
+      jest
+        .spyOn(workspaceRepository, 'findOne')
+        .mockResolvedValueOnce(workspace);
+
+      await service.setupWorkspace(user, 'workspace-id', {
+        name: 'Test Workspace',
+        encryptedMnemonic: 'encryptedMnemonic',
+      });
+
+      expect(teamRepository.addUserToTeam).toHaveBeenCalledWith(
```
sg-gs, line 337: Same here

src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -101,8 +101,49 @@ export class WorkspacesUsecases {
...
+    setupWorkspaceDto: SetupWorkspaceDto,
+  ) {
+    const workspace = await this.workspaceRepository.findOne({
+      id: workspaceId,
+      ownerId: user.uuid,
+      setupCompleted: false,
+    });
+
+    if (!workspace) {
+      throw new NotFoundException('There is no workspace to be setup');
```
sg-gs, line 116: Workspace not found would be better

src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -101,8 +101,49 @@ export class WorkspacesUsecases {
...
+      spaceLimit: BigInt(0),
+      driveUsage: BigInt(0),
+      backupsUsage: BigInt(0),
+      key: setupWorkspaceDto.encryptedMnemonic,
+      deactivated: false,
+      createdAt: new Date(),
+      updatedAt: new Date(),
+    });
+
+    await this.workspaceRepository.addUserToWorkspace(workspaceUser);
```
sg-gs, line 132: If this step fails, we may want to check which step is still failing, so in the case of a massive failure, we can just fix the error and users will only need to retry again. What do you think?

sg-gs (changes requested):

sg-gs (approved):

PR #296, opened 2024-04-16
[PB-1837]: feat/accept invitation to workspace

apsantiso:
This PR adds pre-created user replacement using the same logic we implemented previously for sharings and adds an endpoint to accept workspace invitations.

### When pre-created user keys get replaced?

- When a user registers via normal flow with the invited email.
- When a user registers via a special flow (view) set for external invited users.

### How do users get replaced?

We use the same process as sharings (which is working OK), when the user registers and a new pair of keys are created, we use the default password we set previously to decrypt the private key of the pre-created user and use the new pair of keys to encrypt again the respective encryptionKeys.

### Why did I not add a field to accept the workspace invitation to the pre-created users' registration as we do in sharing?

We had problems with that previously! Sometimes the invitation is not accepted after we update the keys and the userUuid of the invitation! Please reach out to me for a further explanation.

#### Suggestion for this!!

Due to the user uuid in the invitation not being returned correctly if we try to fetch the invitation immediately after it has been updated (it comes with the pre-created user uuid instead of the registered user uuid despite it has been changed a few milliseconds ago). 

We could ask the frontend team to request the accept invite endpoint, this should give the database enough time to process the update of the invitation.

### DIFF
--- a/src/modules/workspaces/repositories/workspaces.repository.spec.ts
+++ b/src/modules/workspaces/repositories/workspaces.repository.spec.ts
@@ -76,16 +76,23 @@ describe('SequelizeWorkspaceRepository', () => {
 
       await repository.bulkUpdateInvitesKeysAndUsers(invites);
 
-      invites.forEach((invite, index) => {
-        expect(workspaceInviteModel.update).toHaveBeenNthCalledWith(
-          index + 1,
-          {
-            invitedUser: invite.invitedUser,
-            encryptionKey: invite.encryptionKey,
-          },
-          { where: { id: invite.id } },
-        );
-      });
+      expect(workspaceInviteModel.update).toHaveBeenNthCalledWith(
+        1,
+        {
+          invitedUser: invites[0].invitedUser,
+          encryptionKey: invites[0].encryptionKey,
+        },
+        { where: { id: invites[0].id } },
+      );
+
+      expect(workspaceInviteModel.update).toHaveBeenNthCalledWith(
+        2,
+        {
+          invitedUser: invites[1].invitedUser,
+          encryptionKey: invites[1].encryptionKey,
+        },
+        { where: { id: invites[1].id } },
+      );
     });
   });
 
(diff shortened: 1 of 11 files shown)


src/modules/workspaces/repositories/workspaces.repository.spec.ts
```diff
@@ -69,6 +70,25 @@ describe('SequelizeWorkspaceRepository', () => {
     });
   });
 
+  describe('bulkUpdateInvitesKeysAndUsers', () => {
+    it('When invites are being updated, then it should update each invite with the correct invitedUser and encryptionKey', async () => {
+      const invites = [newWorkspaceInvite(), newWorkspaceInvite()];
+
+      await repository.bulkUpdateInvitesKeysAndUsers(invites);
+
+      invites.forEach((invite, index) => {
```
sg-gs, line 79: Although this could seem like a great idea, is better to hardcode what we expect to receive rather than doing dynamic checks that imply some logic, as that opens the door to introducing a bug on the test itself.

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -523,6 +523,177 @@ describe('WorkspacesUsecases', () => {
...
+
+      expect(workspaceRepository.addUserToWorkspace).not.toHaveBeenCalled();
+      expect(workspaceRepository.deleteInviteBy).toHaveBeenCalledWith({
+        id: invite.id,
+      });
+    });
+
+    it('When invite is valid, then add user to workspace with respective space limit', async () => {
+      const workspace = newWorkspace();
+      const invite = newWorkspaceInvite({
```
sg-gs, line 574: It could be a good idea to allow the parametrization of the invite's space so we can match it on the `expect` of this test.

sg-gs (changes requested):

sg-gs (approved):

PR #297, opened 2024-04-19
[PB-2049]: feat/added get available and pending to be setup workspaces endpoints

apsantiso:
We need to show users their available workspaces so they can switch between them. We also need to show the current user's pending (to be setup) workspaces. 

This PR intends to add those two endpoints.

- Get available workspaces endpoint: retrieves the already setup workspaces the user belongs to, this endpoint is useful for owners and members of the workspace. 
- Get workspaces pending to be set up: retrieves the current user (thus, the owner of them) workspaces that are yet to be setup.

### DIFF
--- a/src/modules/workspaces/workspaces.usecase.spec.ts
+++ b/src/modules/workspaces/workspaces.usecase.spec.ts
@@ -488,6 +488,89 @@ describe('WorkspacesUsecases', () => {
     });
   });
 
+  describe('getAvailableWorkspaces', () => {
+    it('When trying to get user available workspaces, then workspaces that are already setup should be returned', async () => {
+      const user = newUser();
+      const workspace = newWorkspace({
+        attributes: { setupCompleted: true },
+      });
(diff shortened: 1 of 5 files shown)


src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -488,6 +488,89 @@ describe('WorkspacesUsecases', () => {
...
+      const workspace = newWorkspace({
+        owner,
+        attributes: { setupCompleted: false },
+      });
+
+      jest
+        .spyOn(workspaceRepository, 'findAllBy')
+        .mockResolvedValue([workspace]);
+
+      await service.getWorkspacesPendingToBeSetup(owner);
```
sg-gs, line 565: You are not checking if the returned information is correct

sg-gs (changes requested):

sg-gs (approved):
Please add a description to this PR @apsantiso

PR #298, opened 2024-04-25
[PB-2057]: feat/meet-token-generation

larry-internxt:
It adds an endpoint to get the Meet token to create a video call. It will only work if the email is inside the beta users list.


.env.template
```diff
@@ -41,4 +41,8 @@ STRIPE_SK_TEST=sk_test_y
 GATEWAY_USER=user
 GATEWAY_PASS=gatewaypass
 
-PAYMENTS_API_URL=http://host.docker.internal:8003
\ No newline at end of file
+PAYMENTS_API_URL=http://host.docker.internal:8003
```
sg-gs, line 44: This should be `http://payments-server:8003`, isn't?

src/lib/jitsi.ts
```diff
@@ -0,0 +1,43 @@
...
+      user: {
+        id,
+        name,
+        avatar: '',
+        email,
+        moderator: 'true',
+      },
+      features: {
+        livestreaming: 'true',
+        recording: 'true',
```
sg-gs, line 23: We do not want to enable recording nor transcription nor livestreaming, etc for now. Just enough to do videocalls, nothing else

src/modules/user/user.controller.ts
sg-gs, line 1: Let's use a temporal table for loading the beta users, also, there is no need to use Jitsi naming in all the code base. Let's talk about "Meet" and videocalls better, as they underlying technology could change with the time

sg-gs (changes requested):

sg-gs (approved):

sg-gs:
```

== 20240429103218-create-meet-beta-users-table: migrating =======
== 20240429103218-create-meet-beta-users-table: migrated (0.186s)

✨  Done in 2.41s.
```

PR #303, opened 2024-05-02
[PB-1839] feat/Users can remove themselves from a workspace

jzunigax2:
### Issue
- https://inxt.atlassian.net/browse/PB-1839

Endpoint and useCases to allow workspace members to leave a specific workspace. Per the subtasks on this issue when an owner leaves a workspace a "wipeout" should occur, when a team manager attempts to leave the management of the team(s) is transferred to the workspace owner, and when anyone attempts to leave first file transfer should be enforced.

For the transfer of files I simply perform a check to see whether any WorkspaceItemsUsers record matched the user that is attempting to leave via the createdBy field. 

As this is my first PR all feedback would be appreciated, specially regarding how I handled file and folder deletion when the workspace owner attepts to leave.


src/modules/file/file.usecase.ts
```diff
@@ -309,6 +309,14 @@ export class FileUseCases {
     await this.fileRepository.deleteFilesByUser(user, files);
   }
 
+  /**
+   * Deletes files regardless of the user. Useful in the case of workspace deletion.
+   * @param files Files to be deleted
+   */
+  async delete(files: File[]): Promise<void> {
```
sg-gs, line 316: Enforce the user id usage, as this operation is delicate so we do not want to use the wrong file ids and accidentally delete files that we do not want to delete

src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -721,4 +727,105 @@ export class WorkspacesUsecases {
...
+    const workspace = await this.workspaceRepository.findById(workspaceId);
+
+    if (!workspace) {
+      throw new NotFoundException('Workspace not found');
+    }
+
+    if (!workspace.isUserOwner(user)) {
+      throw new ForbiddenException('You are not the owner of this workspace');
+    }
+    const items =
```
sg-gs, line 744: This is not going to scale. We have a way of handling this: mark the workspace root folder as deleted (`removed=true`). A background process will handle the entire tree deletion for you. No delay and simpler. 

For more info, check [this](https://inxt.atlassian.net/wiki/spaces/T/pages/121733121/DFF+-+v2)

sg-gs (changes requested):

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -2482,5 +2488,178 @@ describe('WorkspacesUsecases', () => {
...
+        jest
+          .spyOn(workspaceRepository, 'findWorkspaceUser')
+          .mockResolvedValue(null);
+
+        await expect(
+          service.leaveWorkspace(workspace.id, user),
+        ).rejects.toThrow(BadRequestException);
+      });
+
+      it('When user is the owner of the workspace, then deleteWorkspace should be called', async () => {
```
sg-gs, line 2556: We should only allow one way of removing the workspace. Is this the only way? If not, we should prevent the owner from leaving the workspace.

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -2482,5 +2488,178 @@ describe('WorkspacesUsecases', () => {
...
+        jest
+          .spyOn(workspaceItemsUsersRepository, 'getAllByUserAndWorkspaceId')
+          .mockResolvedValue([itemUser]);
+
+        await expect(
+          service.leaveWorkspace(workspace.id, user),
+        ).rejects.toThrow(BadRequestException);
+      });
+
+      it('When user is manager of any teams in the workspace then mangement is passed to the owner', async () => {
```
sg-gs, line 2601: _When the user is a manager, then the owner is set as manager of those teams_

Could be a better legend

sg-gs (changes requested):

src/modules/workspaces/workspaces.usecase.spec.ts
jzunigax2: I'm not sure, I would suppose there should be another way to specifically remove the workspace so this flow would be a bit confusing.  I'm now preventing the owner from leaving and left the deleteWorkspace usecase uncalled for the moment, as yes it should probably be executed from a separate endpoint.

src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -1219,4 +1221,82 @@ export class WorkspacesUsecases {
...
+    const rootFolder = await this.folderUseCases.getByUuid(
+      workspace.rootFolderId,
+    );
+
+    await this.folderUseCases.deleteByUser(workspaceUser, [rootFolder]);
+
+    await this.workspaceRepository.deleteById(workspaceId);
+  }
+
+  async leaveWorkspace(
```
sg-gs, line 1252: There is one thing missing here: 
- All the items of that user in its personal workspace should be transferred to the owner's personal folder within that workspace.

sg-gs (changes requested):

src/modules/folder/folder.usecase.ts
```diff
@@ -654,6 +654,34 @@ export class FolderUseCases {
...
+      throw new BadRequestException('Invalid folder name');
+    }
+
+    const newEncryptedName = this.cryptoService.encryptName(
+      newName,
+      folder.parentId,
+    );
+
+    const exists = await this.folderRepository.findByNameAndParentUuid(
+      newEncryptedName,
```
sg-gs, line 668: Search by encrypted and plain name using OR

src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -1364,6 +1364,149 @@ export class WorkspacesUsecases {
...
+      );
+
+    if (
+      !foldersInPersonalRootFolder.length &&
+      !filesInPersonalRootFolder.length
+    )
+      return;
+
+    const workspaceNetworkUser = await this.userRepository.findByUuid(
+      workspace.workspaceUserId,
```
sg-gs, line 1448: Hey, when we talk about the owner of the workspace to set as the owner of the files/folders of a user leaving, we are talking about the _user who pays_, not the special user created for the network. The special user will not be used by anyone to list items, thus, if we transfer things to it, they will be invisible to the user who pays. Isn't that the situation?

sg-gs (changes requested):

src/modules/workspaces/workspaces.usecase.ts
jzunigax2: The special network user is used since the moveFolder useCase expects as an argument the user that owns/created both the folder to be moved and the destination folder. In this case it's the network user who owns the folder to be moved (member's personal root folder) and the destination folder (owner's personal root folder).

This call occurs a few lines below and looks like this:
```typescript
const movedFolder = await this.folderUseCases.moveFolder(
      workspaceNetworkUser,
      memberRootFolder.uuid,
      ownerWorkspaceUser.rootFolderId,
    );
```

sg-gs: That's true @jzunigax2, good catch, I missed that

sg-gs (approved):

PR #305, opened 2024-05-06
[PB-2081]: feat/edit basic workspace details

jzunigax2:
### Issue:
https://inxt.atlassian.net/browse/PB-2081

Endpoint to allow workspace owners to edit basic workspace details (name and description).

### DIFF
--- a/src/modules/workspaces/workspaces.controller.spec.ts
+++ b/src/modules/workspaces/workspaces.controller.spec.ts
@@ -161,12 +161,13 @@ describe('Workspace Controller', () => {
 
       workspacesController.editWorkspaceDetails(workspace.id, user, {
         name: 'new name',
-      }),
-        expect(workspacesUsecases.editWorkspaceDetails).toHaveBeenCalledWith(
-          workspace.id,
-          user,
-          { name: 'new name' },
-        );
+      });
+
+      expect(workspacesUsecases.editWorkspaceDetails).toHaveBeenCalledWith(
+        workspace.id,
+        user,
+        { name: 'new name' },
+      );
     });
   });
   describe('GET /:workspaceId/members', () => {
--- a/src/modules/workspaces/workspaces.controller.spec.ts
+++ b/src/modules/workspaces/workspaces.controller.spec.ts
@@ -163,6 +163,17 @@ describe('Workspace Controller', () => {
         name: 'new name',
       });
 
+      jest
+        .spyOn(workspacesUsecases, 'editWorkspaceDetails')
+        .mockResolvedValue(Promise.resolve());
+
+      await expect(
+        workspacesController.editWorkspaceDetails(workspace.id, user, {
+          name: 'new name',
+          description: 'new description',
+        }),
+      ).resolves.toBeUndefined();
+
       expect(workspacesUsecases.editWorkspaceDetails).toHaveBeenCalledWith(
         workspace.id,
         user,
--- a/src/modules/workspaces/workspaces.controller.spec.ts
+++ b/src/modules/workspaces/workspaces.controller.spec.ts
@@ -159,17 +159,14 @@ describe('Workspace Controller', () => {
       const user = newUser();
       const workspace = newWorkspace({ owner: user });
 
-      await expect(
-        workspacesController.editWorkspaceDetails(workspace.id, user, {
-          name: 'new name',
-        }),
-      ).resolves.toBeTruthy();
-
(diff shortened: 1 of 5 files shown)


src/modules/workspaces/workspaces.controller.ts
```diff
@@ -348,4 +349,24 @@ export class WorkspacesController {
...
   }
+  @Patch('/:workspaceId')
+  @ApiOperation({
+    summary: 'Edit workspace details',
+  })
+  @ApiBearerAuth()
+  @ApiParam({ name: 'workspaceId', type: String, required: true })
+  @UseGuards(WorkspaceGuard)
+  @WorkspaceRequiredAccess(AccessContext.WORKSPACE, WorkspaceRole.OWNER)
+  async editWorkspaceDetails(
```
sg-gs, line 360: If you use `async`, `await` the call. If not, this keyword is not needed

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -299,6 +300,51 @@ describe('WorkspacesUsecases', () => {
...
+    });
+
+    it('When user is not the owner of the workspace, then it should throw', async () => {
+      jest
+        .spyOn(workspaceRepository, 'findById')
+        .mockResolvedValueOnce(workspace);
+
+      await expect(
+        service.editWorkspaceDetails(
+          'workspace-id',
```
sg-gs, line 325: use a uuid

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -299,6 +300,51 @@ describe('WorkspacesUsecases', () => {
...
+          newUser(),
+          editWorkspaceDto,
+        ),
+      ).rejects.toThrow(ForbiddenException);
+    });
+
+    it('When workspace is valid, then it should update the workspace details', async () => {
```
sg-gs, line 332: _When the user is the owner of the workspace, then it should update the workspace details_

Seems to be more accurate

sg-gs (changes requested):

src/modules/workspaces/workspaces.controller.ts
sg-gs: Either return directly and remove async-await or use async-await and store the result in a variable and return afterward

src/modules/workspaces/workspaces.usecase.spec.ts
sg-gs: A uuid is still not being used

sg-gs (changes requested):

src/modules/workspaces/workspaces.controller.spec.ts
```diff
@@ -154,6 +154,22 @@ describe('Workspace Controller', () => {
...
+  describe('PATCH /:workspaceId', () => {
+    it('When workspace details are updated successfully, then it should return.', async () => {
+      const user = newUser();
+      const workspace = newWorkspace({ owner: user });
+
+      workspacesController.editWorkspaceDetails(workspace.id, user, {
+        name: 'new name',
+      });
+
+      expect(workspacesUsecases.editWorkspaceDetails).toHaveBeenCalledWith(
```
sg-gs, line 166: This test is not asserting that the operation has finished successfully unless the method is synchronous, which is not the case. You should do something like: 

`await expect(workspacesController.editWorkspaceDetails(workspace.id, user, { name: 'new name' }).resolves();` or matching the returned content with the injected one by your use case being stubbed.

_And previous to that_, spy the use case to check that, in fact, is has been called with the proper parameters, as a way of improving the test.

src/modules/workspaces/workspaces.usecase.spec.ts
```diff
@@ -313,6 +314,47 @@ describe('WorkspacesUsecases', () => {
...
+    });
+
+    it('When the user is the owner of the workspace, then it should update the workspace details', async () => {
+      jest
+        .spyOn(workspaceRepository, 'findById')
+        .mockResolvedValueOnce(workspace);
+
+      await service.editWorkspaceDetails(workspace.id, user, editWorkspaceDto);
+
+      expect(workspaceRepository.updateBy).toHaveBeenCalledWith(
```
sg-gs, line 349: Same here, you should assert that the method resolves with the expected returned content rather than leaving the test to crash

sg-gs (changes requested):

sg-gs (approved):

PR #308, opened 2024-05-09
feat: modified workspace creation flow

apsantiso:
We need to modify the flow of the workspaces creation. Name and address are now going to be generated with the billing address from Stripe. 

This PR makes optional those fields in the setupWorkspace function. Now we only need the owner's encrypted mnemonic to set up their user on the first access to the workspaces and mark the workspace as ready.

### DIFF
--- a/src/modules/gateway/dto/initialize-workspace.dto.ts
+++ b/src/modules/gateway/dto/initialize-workspace.dto.ts
@@ -1,5 +1,5 @@
 import { ApiProperty } from '@nestjs/swagger';
-import { IsNotEmpty, IsNumber } from 'class-validator';
+import { IsNotEmpty, IsNumber, IsOptional } from 'class-validator';
 
 export class InitializeWorkspaceDto {
   @ApiProperty({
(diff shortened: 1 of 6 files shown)


sg-gs (approved):

src/modules/gateway/dto/initialize-workspace.dto.ts
```diff
@@ -9,6 +9,13 @@ export class InitializeWorkspaceDto {
   @IsNotEmpty()
   ownerId: string;
 
+  @ApiProperty({
+    example: 'Address from billing',
+    description: 'Address of the workspace',
+  })
+  @IsOptional()
+  address?: string;
```
jzunigax2, line 17: In PR #305 I ended up changing the db schema separating the address into its own table since according to issue PB-1556 it seemed the address was more complex than a simple string. That PR hasn't been reviewed yet but this conflicts with those changes.

apsantiso: @jzunigax2 We have not implemented the billing module yet, specially taking into account this is an MVP.

PR #310, opened 2024-05-10
[_] feat: workspaces enable endpoints to retrieve invitations

apsantiso:
This PR exposes 2 new endpoints.

1. Get workspaces/invitations: retrieves current user invitations to any workspace.
2. Get workspaces/:workspaceId/invitations: retrieves pending invitations to the current workspace. Only the owner has access to this endpoint.

### DIFF
--- a/src/modules/workspaces/workspaces.usecase.ts
+++ b/src/modules/workspaces/workspaces.usecase.ts
@@ -716,6 +716,84 @@ export class WorkspacesUsecases {
     return rootFolder;
   }
 
+  async getUserInvites(user: User, limit: number, offset: number) {
+    const invites = await this.workspaceRepository.findInvitesBy(
+      {
+        invitedUser: user.uuid,
+      },
+      limit,
+      offset,
+    );
+
+    const invitesWithWorkspaceData = await Promise.all(
+      invites.map(async (invite) => {
+        const workspace = await this.workspaceRepository.findById(
+          invite.workspaceId,
+        );
+
+        return {
+          ...invite,
+          workspace: workspace.toJSON(),
+        };
+      }),
+    );
+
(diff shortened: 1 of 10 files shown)


src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -419,6 +419,72 @@ export class WorkspacesUsecases {
...
+  ) {
+    const invites = await this.workspaceRepository.findInvitesBy(
+      {
+        workspaceId,
+      },
+      limit,
+      offset,
+    );
+
+    const sharedWithUuids = invites.map((invite) => invite.invitedUser);
```
CandelR, line 460: Why `sharedWithUuids`? Maybe it's because you forgot to rename? If it contains the uuids of the invited users it looks better `invitedUsersUuuids` to me

src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -419,6 +419,72 @@ export class WorkspacesUsecases {
...
+        };
+      }),
+    );
+
+    return invites.map((invite) => ({
+      ...invite,
+      user: usersWithAvatars.find((user) => invite.invitedUser === user.uuid),
+      preCreatedUser: preCreatedUsers
+        .find((user) => invite.invitedUser === user.uuid)
+        ?.toJSON(),
```
CandelR, line 484: What is this precreatedUser field returned for? 
Will we need to differentiate a user from a precreated user to display and manage invitations?

apsantiso: Hey, I merged both precreated users and users so they use the same name. However, the difference between these two is that pre created users are external users  that do not have internxt accounts. I added a boolean field so you can differentiate between external and internal invitations.

PR #312, opened 2024-05-14
[PB-2106]: feat/check if workspace invite is valid

jzunigax2:
Endpoint that provided and inviteId checks for it's existence.


src/modules/workspaces/workspaces.controller.spec.ts
```diff
@@ -240,5 +240,73 @@ describe('Workspace Controller', () => {
...
+          new BadRequestException(errorMessage),
+        );
+
+        try {
+          await workspacesController.validateWorkspaceInvitation(
+            'invalid-id',
+            mockResponse as any,
+          );
+          // If the above line does not throw, the test should fail.
+          fail(
```
apsantiso, line 279: Fail was removed from jest since version 27. You can remove it as it will just throw because "fail is undefined". I do not know why @types/jest still marks this as valid.

src/modules/workspaces/workspaces.controller.spec.ts
```diff
@@ -240,5 +240,73 @@ describe('Workspace Controller', () => {
...
+            'Expected validateWorkspaceInvitation to throw a BadRequestException.',
+          );
+        } catch (error) {
+          // Now we assert that the error is what we expect it to be.
+          expect(error).toBeInstanceOf(BadRequestException);
+          expect(error.response).toEqual({
+            statusCode: 400,
+            message: errorMessage,
+            error: 'Bad Request',
+          });
```
apsantiso, line 289: We do not really need to test the thrown text. You can follow the previous tests and do it  this way

```javascript
await expect(workspacesController.validateWorkspaceInvitation(
            'invalid-id',
            mockResponse)).rejects.toThrow(BadRequestException);
```

src/modules/workspaces/workspaces.controller.spec.ts
```diff
@@ -240,5 +240,73 @@ describe('Workspace Controller', () => {
...
+        workspacesUsecases.validateWorkspaceInvite.mockRejectedValueOnce(
+          new Error('Unexpected error'),
+        );
+
+        const result = await workspacesController.validateWorkspaceInvitation(
+          'id',
+          mockResponse as any,
+        );
+
+        expect(Logger.error).toHaveBeenCalled();
```
apsantiso, line 304: This is not needed. If we remove the logger from there, this test will fail. It is probably that we move the loggers from the controller to an exceptionFilder.

src/modules/workspaces/workspaces.controller.ts
```diff
@@ -84,6 +88,37 @@ export class WorkspacesController {
...
+            error.stack || 'No stack trace'
+          }`,
+        );
+        res.status(HttpStatus.INTERNAL_SERVER_ERROR);
+        errorMessage = 'Internal server error';
+      }
+      return { message: errorMessage };
+    }
+  }
+
```
apsantiso, line 121: I know this type of structure is in another endpoints as well, but it is not needed. Do not use the res to return here, just use the exceptions nestjs provides and throw new internalServerException(), we do not need to write any message or something like that

apsantiso: Preferably, remove the try-catch and just let the request fail, nestjs will do the work. We are going to add an exceptionFilter logger in order for these kind of things get logged correctly, do not worry about logging in this case (unless it is a REALLY special case, we can let the exceptionFilter handle the logging). Also, without the try-catch remember to remove the return await as we do not need to await there anymore to catch the error

apsantiso (changes requested):

apsantiso (approved):

PR #314, opened 2024-05-15
[PB: 2077] Allow owner to edit and add an avatar to workspace

evillalba94:
### Ticket
- https://inxt.atlassian.net/browse/PB-2077

### Update
- Migration to add `avatar` field in workspace
- Update `AvatarService` to configure `Put` & `Delete` options
- Endpoint to `Upload` and `Edit` avatar in workspace
```
URL: /workspaces/:id/avatar
Method: POST
Authorization: Bearer {token}
Body: {
    "file": FILE // Form-data
}
```
- Endpoint to `Delete` avatar in workspace
```
URL: /workspaces/:id/avatar
Method: PUT
Authorization: Bearer {token}
```

### DIFF
--- a/src/externals/avatar/avatar.service.ts
+++ b/src/externals/avatar/avatar.service.ts
@@ -1,13 +1,21 @@
 import { Injectable } from '@nestjs/common';
 import { getSignedUrl } from '@aws-sdk/s3-request-presigner';
-import { GetObjectCommand, S3 } from '@aws-sdk/client-s3';
-
+import {
+  GetObjectCommand,
+  ObjectCannedACL,
+  S3Client,
+  PutObjectCommand,
+  DeleteObjectCommand,
+} from '@aws-sdk/client-s3';
+import { Express } from 'express';
 import { User } from '../../modules/user/user.domain';
+import { Workspace } from '../../modules/workspaces/domains/workspaces.domain';
+import { v4 } from 'uuid';
 
 @Injectable()
 export class AvatarService {
-  async getDownloadUrl(avatar: User['avatar']): Promise<string> {
-    const s3 = new S3({
+  AvatarS3Instance(): S3Client {
+    return new S3Client({
       endpoint: process.env.AVATAR_ENDPOINT,
       region: process.env.AVATAR_REGION,
       credentials: {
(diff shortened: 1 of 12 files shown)


src/externals/avatar/avatar.service.ts
```diff
@@ -35,4 +48,32 @@ export class AvatarService {
...
+      ACL: ObjectCannedACL.public_read,
+      ContentType: file.mimetype,
+    };
+
+    const putObjectCommand = new PutObjectCommand(putObjectInput);
+    await s3Client.send(putObjectCommand);
+    return keyObject;
+  }
+
+  async deleteAvatar(avatarKey: User['avatar'] | Workspace['avatar']) {
```
sg-gs, line 70: set the return type for this function

src/modules/workspaces/workspaces.controller.spec.ts
```diff
@@ -241,4 +241,65 @@ describe('Workspace Controller', () => {
...
+      await workspacesController.uploadAvatar(file, workspace.id);
+      expect(workspacesUsecases.upsertAvatar).toHaveBeenCalledWith(
+        workspace.id,
+        file,
+      );
+    });
+  });
+
+  describe('DELETE /:workspaceId/avatar', () => {
+    it('When workspaceId is null, throw error.', async () => {
```
sg-gs, line 288: _When workspaceId is null, then it fails_

src/modules/workspaces/workspaces.controller.spec.ts
```diff
@@ -241,4 +241,65 @@ describe('Workspace Controller', () => {
...
+      const workspaceId = null;
+      await expect(
+        workspacesController.deleteAvatar(workspaceId),
+      ).rejects.toThrow('Invalid workspace ID');
```
sg-gs, line 292: We should throw an error not an error message

sg-gs (changes requested):

src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -1248,4 +1265,79 @@ export class WorkspacesUsecases {
...
+    if (workspace.avatar) {
+      try {
+        await this.avatarService.deleteAvatar(workspace.avatar);
+      } catch (err) {
+        Logger.error(
+          `[WORKSPACE/SET_AVATAR]Deleting the avatar in workspaceId: ${
+            workspace.id
+          } has failed. Error: ${(err as Error).message}`,
+        );
+        throw new Error('Deleting avatar to workspace has failed');
```
apsantiso, line 1298: Just propagate the error please. So we can catch/log later easily when we add a proper filterException. Nestjs will just show an internal server error unless you specify a specific HTTP error, so not worry about this error being parsed to the response.

apsantiso (changes requested):

apsantiso:
@sg-gs Done. I made a few comments about labeling (required) and some tests that could be done differently (although they don't really need to be changed).

The upload without buffering is working flawlessly. Great job!

apsantiso:
@evillalba94 Sorry for being intrusive, but I merged something into the main branch and made sure it was merged into this branch too. It is related to stop using bigints so we do not need to add toJSON  everytime to prevent nestjs from failing

evillalba94:
@apsantiso 
Ok no problem, by the way, thanks for the review and comments.

apsantiso (approved):

PR #317, opened 2024-05-21
[PB-2144] feat: share files and folders with other members in the workspace

apsantiso:
### Description

This PR allow users in the workspace to share files and folders with other members in the workspace.

1. **Migration Scripts:**
   - Removed `sharings` constraint.
   - Added `sharedWithType` column (enum).

2. **Domain Modifications:**
   - Updated Sharing domain with `sharedWithType`.
   - Added `SharingActionName` to permissions domain.

3. **Repository Changes:**
   - Implemented `findPermissions` for sharings.
   - Added `canUserPerformActionInSharing` use case.

5. **Guards and Decorators:**
   - Created `SharingPermissionsGuard` for checking permissions in shared items.
   - Added `BehalfUserDecorator` to retrieve behalfUser from requests.

6. **Controller Updates:**
   - Applied `SharingPermissionsGuard` to `createFile` endpoint.
   - Added `shareItemWithMember`, `getFoldersInsideSharedFolder`, and `getFilesInsideSharedFolder` endpoints to WorkspaceController.

7. **Miscelaneous Adjustments:**
   - Moved RoleModel to resolve TypeScript relation issues.

### DIFF
--- a/src/modules/sharing/guards/sharing-permissions.guard.ts
+++ b/src/modules/sharing/guards/sharing-permissions.guard.ts
@@ -121,7 +121,7 @@ export class SharingPermissionsGuard implements CanActivate {
       this.workspaceUseCases.findUserInTeam(requester.uuid, teamId),
     ]);
 
-    return userIsAllowedToPerfomAction && !!isUserPartOfTeam;
+    return userIsAllowedToPerfomAction && !!isUserPartOfTeam?.teamUser;
   }
 
   async isUserAbleToPerfomAction(
--- a/src/modules/sharing/guards/sharing-permissions.guard.ts
+++ b/src/modules/sharing/guards/sharing-permissions.guard.ts
@@ -103,7 +103,7 @@ export class SharingPermissionsGuard implements CanActivate {
     action: SharingActionName,
   ) {
     const [userIsAllowedToPerfomAction, isUserPartOfTeam] = await Promise.all([
-      this.sharingUseCases.canPerfomActionInWorkspace(
+      this.sharingUseCases.canPerfomAction(
         requester.uuid,
         sharedRootFolderId,
         action,
--- a/src/modules/sharing/guards/sharing-permissions.guard.ts
+++ b/src/modules/sharing/guards/sharing-permissions.guard.ts
@@ -108,4 +108,20 @@ export class SharingPermissionsGuard implements CanActivate {
 
     return userIsAllowedToPerfomAction;
   }
+
+  async isIndividualUserAbleToPerfomAction(
+    requester: User,
+    sharedRootFolderId: Folder['uuid'],
+    action: SharingActionName,
+  ) {
+    const userIsAllowedToPerfomAction =
+      await this.sharingUseCases.canUserPerformActionInSharing(
+        requester.uuid,
+        sharedRootFolderId,
+        action,
+        SharedWithType.Individual,
+      );
+
+    return userIsAllowedToPerfomAction;
+  }
 }
--- a/src/modules/sharing/guards/sharing-permissions.guard.ts
+++ b/src/modules/sharing/guards/sharing-permissions.guard.ts
@@ -71,13 +71,22 @@ export class SharingPermissionsGuard implements CanActivate {
       throw new ForbiddenException('Invalid token');
     }
 
-    const userIsAllowedToPerfomAction =
-      await this.isTeamMemberAbleToPerformAction(
+    let userIsAllowedToPerfomAction = false;
+
+    if (decoded.workspace) {
+      userIsAllowedToPerfomAction = await this.isTeamMemberAbleToPerformAction(
         requester,
         decoded.workspace.teamId,
         decoded.sharedRootFolderId,
         action,
       );
+    } else {
+      userIsAllowedToPerfomAction = await this.isUserAbleToPerfomAction(
+        requester,
+        decoded.sharedRootFolderId,
+        action,
+      );
+    }
 
     if (!userIsAllowedToPerfomAction) {
       return false;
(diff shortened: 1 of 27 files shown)


src/modules/sharing/guards/sharing-permissions.guard.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (typeof decoded === 'string') {
+      throw new ForbiddenException('Invalid token');
+    }
+
+    const userIsAllowedToPerfomAction =
+      await this.isWorkspaceMemberAbleToPerformAction(
+        requester,
+        decoded.sharedRootFolderId,
+        action,
+      );
```
apsantiso, line 77: I only added support for sharing between members, but we can adapt this guard so it works with individual sharings and teams in a future. My thought is use the decoded token to know if the sharing the user is navigating through is for individuals, teams or members and make the respective checks according to that.

src/modules/sharing/models/index.ts
```diff
@@ -3,39 +3,44 @@ import {
...
 import { FolderModel } from '../../folder/folder.model';
 import { FileModel } from '../../../modules/file/file.model';
 import { PreCreatedUserModel } from '../../../modules/user/pre-created-users.model';
+import { WorkspaceTeamModel } from '../../workspaces/models/workspace-team.model';
+import { SharingRolesModel } from './sharing-roles.model';
 
 @Table({
   underscored: true,
   timestamps: true,
-  tableName: 'permissions',
```
apsantiso, line 25: This may be seen as I changed this table, but actually I just switched the table roles and Permissions position because typescript complains about RoleModel not being defined yet.

src/modules/workspaces/workspaces.controller.ts
```diff
@@ -339,15 +347,106 @@ export class WorkspacesController {
...
+    if (!behalfUser) {
+      throw new ForbiddenException('You have not permissions to do this!');
+    }
+
+    return this.workspaceUseCases.createFile(
+      behalfUser,
+      workspaceId,
+      createFileDto,
+    );
+  }
```
apsantiso, line 368: We need to add logs in a future. When that time comes, we can bring back the @UserDecorator to get the actual requester and log from the controller (what I think is going to be the right way to do it).

src/modules/workspaces/workspaces.controller.ts
```diff
@@ -339,15 +347,106 @@ export class WorkspacesController {
...
+    @Body() ShareItemWithMemberDto: ShareItemWithMemberDto,
+  ) {
+    return this.workspaceUseCases.shareItemWithMember(
+      user,
+      workspaceId,
+      ShareItemWithMemberDto,
+    );
+  }
+
+  @Get(':workspaceId/items/:sharedFolderId/folders')
```
apsantiso, line 394: It would be good if someone can help me with a better naming for this :)

src/modules/sharing/guards/sharing-permissions.guard.ts
```diff
@@ -0,0 +1,111 @@
...
+    return true;
+  }
+
+  async isWorkspaceMemberAbleToPerformAction(
+    requester: User,
+    sharedRootFolderId: Folder['uuid'],
+    action: SharingActionName,
+  ) {
+    const userIsAllowedToPerfomAction =
+      await this.sharingUseCases.canUserPerformActionInSharing(
```
sg-gs, line 102: Perfect! Keeping the business logic on the use cases even here is a beast!

src/modules/workspaces/workspaces.controller.ts
sg-gs: That's a great approach

src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -721,6 +734,298 @@ export class WorkspacesUsecases {
...
+
+    const team = await this.teamRepository.getTeamById(
+      shareWithTeamDto.sharedWith,
+    );
+
+    if (!team) {
+      throw new BadRequestException('Team is not valid');
+    }
+
+    const [existentSharing, workspace] = await Promise.all([
```
sg-gs, line 766: Ensure the workspace still exists

sg-gs (changes requested):

src/modules/sharing/sharing.service.spec.ts
```diff
@@ -395,10 +400,305 @@ describe('Sharing Use Cases', () => {
...
+      );
+
+      expect(result).toEqual(
+        expect.objectContaining({
+          folders: expect.arrayContaining([
+            expect.objectContaining({
+              sharingId: sharing.id,
+              encryptionKey: sharing.encryptionKey,
+              dateShared: sharing.createdAt,
+              sharedWithMe: user.uuid !== sharing.folder.user.uuid,
```
sg-gs, line 639: Set the value directly, not a logical comparison. Otherwise, this line would need to be tested also, which is what we want to avoid in a test.

src/modules/workspaces/workspaces.controller.ts
```diff
@@ -458,15 +470,165 @@ export class WorkspacesController {
...
+    return this.sharingUseCases.getSharedFoldersInWorkspace(
+      user,
+      teamId,
+      page,
+      perPage,
+      order,
+    );
+  }
+
+  @Get(':workspaceId/:teamId/shared/:sharedFolderId/folders')
```
sg-gs, line 568: Following REST conventions, it needs to be like this:

GET /workspaces/:workspaceId/**teams**/:teamId/shared/:sharedFolderId/folders

src/modules/workspaces/workspaces.controller.ts
```diff
@@ -458,15 +470,165 @@ export class WorkspacesController {
...
+      teamId,
+      user,
+      sharedFolderId,
+      WorkspaceItemType.Folder,
+      token,
+      { page, perPage, order },
+    );
+  }
+
+  @Get(':workspaceId/:teamId/shared/:sharedFolderId/files')
```
sg-gs, line 601: Same here

sg-gs (changes requested):

PR #325, opened 2024-06-04
[PB-2218]: feat/add-folders-depth

larry-internxt:
- Adds an indexed depth column to folders
- Adds function to calculate folders depth by its parent depth
- Adds function to recursively update all child depths if a parent is updated
- Adds trigger before Create/Update folder: calculate its depth by using the parent
- Adds trigger after Update folder: recalculate and fix all the child depths


src/modules/folder/folder.domain.ts
```diff
@@ -34,6 +34,7 @@ export class Folder implements FolderAttributes {
   updatedAt: Date;
   size: number;
   sharings?: Sharing[];
+  depth: number | null;
```
apsantiso, line 37: Could you please modify this type to be `depth?: number` instead? This would prevent us from adding `depth: null` to every instance where Folder.build({...values}) is used. This field will be automatically populated by the database triggers, so having it as undefined or null has the same effect unless we are specifically comparing undefined and null.

I bring this up because, for instance, you had to modify some unit tests due to this type, and I would also need to add many depth: null entries in the workspaces branch.

apsantiso (approved):

apsantiso (dismissed):

PR #327, opened 2024-06-10
feat: calculate member usage in workspace

apsantiso:
Due to the current files and folders structure, and the files being removed using a background process when the parent folder is deleted. We need a way to calculate the workspace member's usage on the fly. 

However, using a sum of all the files with TRASHED or EXISTS status would affect the scalability of the request, so we need to manage it using  a sort of "caching" with the last_sync_at date.  This it is just a temporal solution and it should be modified once the deltas solution is implemented.

### DIFF
--- a/src/modules/file/file.repository.ts
+++ b/src/modules/file/file.repository.ts
@@ -49,6 +49,13 @@ export interface FileRepository {
     folderUuid: FileAttributes['folderUuid'],
     status: FileAttributes['status'],
   ): Promise<File | null>;
+  getSumSizeOfTrashedAndExistentFiles(
+    createdBy: WorkspaceItemUserAttributes['createdBy'],
+    workspaceId: WorkspaceAttributes['id'],
+    limit: number,
+    offset: number,
+    order: Array<[keyof File, string]>,
+  ): Promise<{ size: string }[]>;
   updateByFieldIdAndUserId(
     fileId: FileAttributes['fileId'],
     userId: FileAttributes['userId'],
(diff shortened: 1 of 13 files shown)


src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -450,6 +450,90 @@ export class WorkspacesUsecases {
...
+    const member = await this.workspaceRepository.findWorkspaceUser({
+      memberId: user.uuid,
+      workspaceId,
+    });
+
+    if (!member) {
+      throw new BadRequestException('User not valid');
+    }
+
+    const syncStartedAt = new Date();
```
apsantiso, line 466: Get the date from here to prevent cases such as some file being set as DELETED while this is running. 

This could happen if the user deletes a very nested folder and children files are still being updated by the trigger in the database. 

The next time the sync is ran, it should include the files that were not calculated in this call.

src/modules/file/file.repository.ts
```diff
@@ -292,6 +304,53 @@ export class SequelizeFileRepository implements FileRepository {
...
+      include: {
+        model: WorkspaceItemUserModel,
+        attributes: [],
+        where: { createdBy, workspaceId, ...workspaceItemDateFilters },
+      },
+      order: options?.order,
+    });
+
+    return sizes as unknown as { size: string }[];
+  }
```
apsantiso, line 352: I wanted to use .sum or try to get the total using the raw:true option with sequelize.literal inside the attributes field. 
However, sequelize is not very happy when using includes, group by and a SUM inside attributes at the same time. It continues asking to group by fields that just makes the query complex, harder to read and it still returns an array.

The only way to get the sum directly from the DB is probably using a raw query, but in this case, I preferred not to use that approach and just use .reduce to sum the values.

src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -450,6 +450,90 @@ export class WorkspacesUsecases {
...
+          [dateField]: dateFrom,
+        },
+      );
+
+      filesFetched = sizesChunk.length;
+
+      const filesSize = sizesChunk.reduce(
+        (sum, file) => sum + Number(file.size),
+        0,
+      );
```
apsantiso, line 528: Check https://github.com/internxt/drive-server-wip/pull/327/files#r1633866492

PR #334, opened 2024-06-17
[PB-557] fix/file size as number

jzunigax2:
Similar to #321, file size's type is changed from BigInt to Number. This would be a breaking change for the mac desktop client, but `convertSizeMiddleware` is applied to problematic routes as reported by @PixoDev in order to convert the numeric size into string. These routes are:
- PUT `/files/:fileUUID`
- GET `/files/:fileUUID/meta`
- GET `/folders/:folderId/files`
- GET `/files?updatedAt=ISODate`


sg-gs (dismissed):

jzunigax2:
Unable to merge until we can confirm this doesn't break web and mobile clients @CandelR / @PixoDev

apsantiso:
This has been stalled for a long time. Let's close it and it could be reopened when needed.

PR #341, opened 2024-06-25
[PB-2284]: Feat/register notification token endpoint

jzunigax2:
Added `user_notification_tokens` table and an endpoint to register these device tokens. Currently, this is aimed at implementing Apple's push notification service for the macOS client to signal real-time changes. Still, the new table contains a type property in case this could be used for a different purpose in the future.

### DIFF
--- a/migrations/20240624200747-create-user-tokens-table.js
+++ b/migrations/20240624200747-create-user-tokens-table.js
@@ -5,39 +5,49 @@ const tableName = 'user_notification_tokens';
 /** @type {import('sequelize-cli').Migration} */
 module.exports = {
   async up(queryInterface, Sequelize) {
-    await queryInterface.createTable(tableName, {
-      id: {
-        type: Sequelize.UUID,
-        primaryKey: true,
-        defaultValue: Sequelize.UUIDV4,
-      },
-      user_id: {
-        type: Sequelize.STRING(36),
-        allowNull: false,
-        references: {
-          model: 'users',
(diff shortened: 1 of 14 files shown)


sg-gs (changes requested):
This PR is not handling three things related to the security and abuse prevention of this API: 
- Throttling requests on that endpoint
- Removal process of those tokens: do they expire or something?
- Setting a max number of tokens per user: we do not want to allow 20 devices to be connected. Up to 10 is fine for now.

@PixoDev @jzunigax2

jzunigax2:
> This PR is not handling three things related to the security and abuse prevention of this API:
> 
>     * Throttling requests on that endpoint
> 
>     * Removal process of those tokens: do they expire or something?
> 
>     * Setting a max number of tokens per user: we do not want to allow 20 devices to be connected. Up to 10 is fine for now.
> 
> 
> @PixoDev @jzunigax2

It seems Apple's tokens don't have an expiration date. However, when sending a notification the API could respond with 410, indicating that the token is no longer active due to a variety of reasons. I'll make sure to handle this case when integrating the APN service.

@sg-gs

sg-gs (approved):

sg-gs:
Migration run @jzunigax2 

```
== 20240624200747-create-user-tokens-table: migrating =======
== 20240624200747-create-user-tokens-table: migrated (0.352s)

✨  Done in 2.52s
```

PR #359, opened 2024-07-12
[PB-2309] feat: created endpoint for folder and files existence

apsantiso:
The frontend loops through all the files inside a folder to know if there is a file with the same name already created inside it. This operation is a burden for the backend and not eficient.


apsantiso:
I am going to force merge this PR to allow the frontend team to continue working on production without having to start from scratch with all the necessary workspaces, files, etc., as they would if we deployed a new environment for this PR. Let's address the fixes and changes after the review in another PR. @sg-gs

PR #388, opened 2024-08-12
[PB-2504: fix/file transfer space validation

jzunigax2:
Temporary check to validate the owner has enough space to receive a removed member's files. This is temorary as it should be replaced by a "transfer ownership"  feature in an upcoming release.

### DIFF
--- a/src/modules/workspaces/workspaces.usecase.ts
+++ b/src/modules/workspaces/workspaces.usecase.ts
@@ -2182,6 +2182,33 @@ export class WorkspacesUsecases {
       throw new ForbiddenException('You are the owner of this workspace');
     }
 
+    const memberUsage = await this.calculateFilesSizeSum(
+      user.uuid,
+      workspace.id,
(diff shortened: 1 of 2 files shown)


src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -2203,6 +2203,33 @@ export class WorkspacesUsecases {
...
+    const memberUsage = await this.calculateFilesSizeSum(
+      user.uuid,
+      workspace.id,
+      [FileStatus.EXISTS],
+    );
+
+    const ownerUsage = await this.calculateFilesSizeSum(
+      workspace.ownerId,
+      workspace.id,
+      [FileStatus.EXISTS],
```
sg-gs, line 2215: EXISTS + TRASH should be the calculus @jzunigax2

sg-gs (changes requested):

sg-gs (approved):

PR #401, opened 2024-09-30
[PB-3009]: feat/creation modification times on folders

jzunigax2:
Added `creationTime` and `modificationTime` fields to the folders table. When creating a new folder it is allowed to set these parameters optionally.
Also when modifying a folder or file's name the respective `modificationTime` field gets updated.


migrations/20240929040050-create-modificationtime-index-on-files.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const tableName = 'files';
+const indexName = 'files_modification_time_index';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
```
sg-gs, line 8: I need these indexes to be created CONCURRENTLY @jzunigax2. Thus, I can run them in the background without locking the `folders` table. It's less clean but far more effective.

sg-gs: Ref: https://www.postgresql.org/docs/current/sql-createindex.html

migrations/20240929044035-create-modificationtime-index-on-folders.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const tableName = 'folders';
+const indexName = 'folders_modification_time_index';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.addIndex(tableName, ['modification_time'], {
```
sg-gs, line 9: Same here

apsantiso (approved):
LGTM, let's just wait for the migrations to be added.

sg-gs:
Migrations run @jzunigax2 @apsantiso. Ready to merge.

apsantiso:
@sg-gs @jzunigax2 Just made a small commit to remove duplicated migrations to prevent our locals from breaking.

PR #403, opened 2024-10-01
[IN-79]: fix/cleanup-files-name-indexes

sg-gs:
This PR aims to reduce the infra debt related to the files field `name`, which is causing multiple performance problems while not giving advantages of any type. 

The migrations fix the cases where files were being duplicated by name due to not migrating the index where `status = 'EXISTS'` during the transition we did from `deleted`/`removed` fields to `status`.


sg-gs:
The migrations have been already run succesfully.

PR #407, opened 2024-10-07
[PB-3045]: fix/get-ancestors-in-workspace

jzunigax2:
Added a function that exclude direct children of a root folder, in the case of workspaces those are indiviudual root folder. So this fixes an error where currently the ancestors endpoint returns the user's root folder.

### DIFF
--- a/migrations/20241004142347-get-folder-ancestor-omit-root-children.js
+++ b/migrations/20241004142347-get-folder-ancestor-omit-root-children.js
@@ -1,0 +1,42 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+      CREATE OR REPLACE FUNCTION get_folder_ancestors_excluding_root_children(folder_id UUID, u_id INT)
+      RETURNS setof folders AS $$
+      BEGIN
+        RETURN QUERY
+        WITH RECURSIVE hier AS (
+          SELECT c.*
+          FROM folders c
+          WHERE c.removed = FALSE
+          AND c.uuid = folder_id
+          UNION
+          SELECT f.*
+          FROM folders f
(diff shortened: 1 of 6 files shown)


jzunigax2:
I added 35k folders with one being the parent of the next one, and ran both the `get_folder_ancestors ` and `get_folder_ancestors_excluding_root_children` functions.

`get_folder_ancestors`:
|QUERY PLAN|
|----------|
|Function Scan on get_folder_ancestors  (cost=0.25..10.25 rows=1000 width=1250) (actual time=261.760..269.839 rows=35044 loops=1)|
|Planning Time: 0.221 ms|
|Execution Time: 289.829 ms|

`get_folder_ancestors_excluding_root_children`:
|QUERY PLAN|
|----------|
|Function Scan on get_folder_ancestors_excluding_root_children  (cost=0.25..10.25 rows=1000 width=1250) (actual time=229.367..237.720 rows=35044 loops=1)|
|Planning Time: 0.153 ms|
|Execution Time: 243.699 ms|

sg-gs:
Migration run @jzunigax2 

```
== 20241004142347-get-folder-ancestor-omit-root-children: migrating =======
== 20241004142347-get-folder-ancestor-omit-root-children: migrated (0.236s)

✨  Done in 3.97s.
```

sg-gs (approved):

apsantiso:
Hey, @jzunigax2 @edisonjpadilla, for these kind of cases, instead of adding a query param, it is better to attach something to the request using the workspaces guard so we can control what are we going to call instead of rely on the frontend.

You can check the guard here: https://github.com/internxt/drive-server-wip/blob/09b6dcb8ae0f210691e359810d733a144bd96058/src/modules/workspaces/guards/workspaces-resources-in-items-in-behalf.guard.ts#L107

We can add the workspaceId or even the workspace object to the request object and get it via a decorator like we do with the user via the authorization guard. Please try to use the guards as much as possible if possible, so we can prevent security issues

PR #411, opened 2024-10-18
[_]: fix/retrun user's root folder when retrieving ancestors in workspaces

jzunigax2:
In #407 I excluded both the workspace root folder and the user's root folder from the ancestors result. Turns out that that the ancestors call for personal drives always returns the root folder, so for compatibility this PR fixes this and now only the workspace's root folder is excluded.


sg-gs:
Migration run

```

== 20241018040841-modify-get-folder-ancestor-omit-root-children-function: migrating =======
== 20241018040841-modify-get-folder-ancestor-omit-root-children-function: migrated (0.172s)

✨  Done in 2.83s.
```

sg-gs (approved):

PR #417, opened 2024-10-29
[PB-1783]: feat/add-type-to-update-file

larryrider:
It adds the ability to edit the extension of the files


src/modules/file/dto/update-file-meta.dto.ts
```diff
@@ -1,11 +1,20 @@
...
+  plainName?: string;
+
+  @IsString()
+  @ApiProperty({
+    example: 'New type',
+    description: 'The new type that the file is going to have',
+  })
+  @IsOptional()
+  type?: string;
 }
```
apsantiso, line 20: since you are using .trim() in the service, you could move that to this DTO.

```
import { ApiProperty } from '@nestjs/swagger';
import { IsOptional, IsString, IsNotEmpty } from 'class-validator';
import { Transform } from 'class-transformer';

export class UpdateFileMetaDto {
  @IsOptional()
  @IsString()
  @IsNotEmpty({ message: 'Filename cannot be empty' })
  @Transform(({ value }) => value?.trim())
  @ApiProperty({
    example: 'New name',
    description: 'The name the file is going to be updated to',
  })
  plainName?: string;

  @IsOptional()
  @IsString()
  @IsNotEmpty({ message: 'File type cannot be empty' })
  @Transform(({ value }) => value?.trim())
  @ApiProperty({
    example: 'New type',
    description: 'The new type that the file is going to have',
  })
  type?: string;
}
```

src/modules/file/file.controller.ts
```diff
@@ -213,6 +213,9 @@ export class FileController {
     @Body() updateFileMetaDto: UpdateFileMetaDto,
     @Client() clientId: string,
   ) {
+    if (!updateFileMetaDto) {
+      throw new BadRequestException('Missing update file metadata');
+    }
```
apsantiso, line 218: A small tip, you can use Nestjs tools to strip non desired values from the body of the request, allowing you to check if at least one of the fields is present by using `Object.keys(updateFileMetaDto).length === 0 `

  ```
@UsePipes(
    new ValidationPipe({
      whitelist: true,
    }),
  )
  async updateFileMetadata(
    @UserDecorator() user: User,
    @Param('uuid', ValidateUUIDPipe)
    fileUuid: File['uuid'],
    @Body() updateFileMetaDto: UpdateFileMetaDto,
    @Client() clientId: string,
  ) {
    if (!updateFileMetaDto || Object.keys(updateFileMetaDto).length === 0) {
      throw new BadRequestException('Missing update file metadata');
    }
```

File: src/modules/file/file.usecase.ts [hunk 3]
```diff
@@ -155,32 +156,77 @@ export class FileUseCases {
...
+    const emptyType =
+      newFileMetadata.type === null ||
+      newFileMetadata.type?.trim().length === 0;
+
+    if (emptyPlainName && emptyType) {
+      throw new BadRequestException('Filename can not be empty');
+    }
+    return { metadataHasPlainName, metadataHasType, emptyPlainName, emptyType };
+  }
+
```
apsantiso, line 175: Using the tools I left in other comments, I think you could get rid of this function (just take care of the type: null case sent in the body). 

Not a bad idea to re-check things in services, but If we have already class validator and some tools nestjs provides us with to check things beforehand, we can reduce code in services. More like a personal opinion tho, so take it with a grain of salt!

apsantiso (dismissed):
Just left some comments, feel free to use what you feel it could be useful, since the implementation works, it is a personal opinion and not a "this is bad" review.

src/modules/file/file.usecase.ts
```diff
@@ -155,32 +156,77 @@ export class FileUseCases {
...
+      // plainName is provided, so the new name has to be encrypted
+      plainName = newFileMetadata.plainName || '';
+      cryptoFileName = this.cryptoService.encryptName(plainName, file.folderId);
+    } else {
+      // plainName is not provided, so the name remains unchanged
+      plainName =
+        file.plainName ??
+        this.cryptoService.decryptName(file.name, file.folderId);
+      cryptoFileName = file.name;
+    }
```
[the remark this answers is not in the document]
larryrider: I need to change some things to do it like this, but it can be a better idea 👍

src/modules/file/file.usecase.ts
larryrider: yeah it sounds good to me, but i will still use the empty checks on the service just because we need to ensure that after looking into the db the update is not done with an empty filename (empty name + empty extension)

src/modules/file/dto/update-file-meta.dto.ts
larryrider: mmm not really as the trim() is done just to ckeck if its empty, not for the update property itself

src/modules/file/file.controller.ts
larryrider: I have moved this to the service as its critical for the good functioning of the usecase by itself

sg-gs: That is not correct @larryrider. The responsibility of the controller is precisely to validate the input of the request before passing it to the service/usecase. 

If you have business rules regarding to what can be updated, what not and in which situation, use the File entity to add the business rules to the domain, so you keep those details where they should be. 

Otherwise, a rule that is bound to the file and should be enforced always, it's enforced per each use case, increasing the margin to miss that detail in other usecases. 

Then, we can just call that method (like `isValidPlainName()`) on the beginning of the use case method.

src/modules/file/file.usecase.ts
```diff
@@ -155,32 +159,68 @@ export class FileUseCases {
...
       status: FileStatus.EXISTS,
+      userId: user.id,
     });
+    if (!file) {
+      throw new NotFoundException('File not found');
+    }
 
-    if (!file.isOwnedBy(user)) {
-      throw new ForbiddenException('This file is not yours');
+    let plainName =
```
sg-gs, line 191: For instance, `isOwnedBy` is a method thought like that. Here, you are increasing the code complexity of a context that should be centralised in the file entity. If you have any other situation where this method cannot be reused, you will need copy-paste all these checks.

larryrider: In this case I didnt add checks for the owner, I just removed that isOwnedBy check because I added the userId to the file.findOneBy, trying to speed the SQL petition by adding the owner filter directly to the query.
But if there is no improvement, I can revert it if it helps to the context

sg-gs: Updating the metadata, like the name or the type is not an operation that could be a reason to complain about speeds, hence, one select more or less is not worth it if it is at the expense of making the code less clear or hiding the business rules or the security checks

src/modules/file/file.usecase.ts
```diff
@@ -158,29 +160,51 @@ export class FileUseCases {
...
+      newFileMetadata.plainName ??
+      file.plainName ??
+      this.cryptoService.decryptName(file.name, file.folderId);
+    if (isStringEmpty(plainName)) {
```
sg-gs, line 185: Please, enforce these checks using the File entity rather than keeping all the business rules exposed here

sg-gs (changes requested):

sg-gs (approved):

sg-gs (approved):

PR #422, opened 2024-11-05
[PB-3097, PB-2977] Filter files in trash - Delete files from deleted folders

evillalba94:
### Ticket
- https://inxt.atlassian.net/browse/PB-3097

### Updates
- Fixed the files to be displayed in the `Recents` section
- Improved the handling of files in the `Trash` section

### DIFF
--- a/src/modules/file/file.usecase.ts
+++ b/src/modules/file/file.usecase.ts
@@ -36,7 +36,6 @@ import { WorkspaceAttributes } from '../workspaces/attributes/workspace.attribut
 import { Folder } from '../folder/folder.domain';
 import { getPathFileData } from '../../lib/path';
 import { isStringEmpty } from '../../lib/validators';
-import { Op } from 'sequelize';
 
 export type SortParamsFile = Array<[SortableFileAttributes, 'ASC' | 'DESC']>;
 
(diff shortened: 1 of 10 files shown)


src/modules/folder/folder.usecase.ts
```diff
@@ -810,6 +813,24 @@ export class FolderUseCases {
...
+      deleted: false,
+      deletedAt: null,
+      status: FileStatus.EXISTS,
+    };
+
+    await this.fileUsecases.updateManyByFileIdAndUserId(
+      fileIds,
+      user.id,
+      updateFileData,
+    );
```
apsantiso, line 832: It is better to make a query that updates all the files inside that folder instead of fetching and updating them. We have no limits for files inside a folder, so if an user has 10k files in a folder, those are 10k files you're fetching/looping through here.

src/modules/file/file.usecase.ts
```diff
@@ -36,6 +36,7 @@ import { WorkspaceAttributes } from '../workspaces/attributes/workspace.attribut
 import { Folder } from '../folder/folder.domain';
 import { getPathFileData } from '../../lib/path';
 import { isStringEmpty } from '../../lib/validators';
+import { Op } from 'sequelize';
```
apsantiso, line 39: Could you try to use this in the repository? You can find the reason in this comment: https://github.com/internxt/drive-server-wip/pull/428#discussion_r1837767776

apsantiso (changes requested):

sg-gs:
This PR it's non-sense. The issue with this thing is related to the background deletion system we have on Drive. It is deleted eventually. If you want to filter _Recents_ and _Trash_ without still-not-deleted items, you need to do a recursive query when listing those items, to check if any of their folder parents `removed` field is `true` @apsantiso @evillalba94

PR #428, opened 2024-11-08
[_]: chore/refactor accept invite

jzunigax2:
Add transaction support to acceptWorkspaceInvite usecase, to ensure atomicity and data integrity.

### DIFF
--- a/src/modules/folder/folder.usecase.ts
+++ b/src/modules/folder/folder.usecase.ts
@@ -30,6 +30,7 @@ import { WorkspaceAttributes } from '../workspaces/attributes/workspace.attribut
 import { FileUseCases } from '../file/file.usecase';
 import { File, FileStatus } from '../file/file.domain';
 import { CreateFolderDto } from './dto/create-folder.dto';
+import { Transaction } from 'sequelize';
 
 const invalidName = /[\\/]|^\s*$/;
 
@@ -355,6 +356,7 @@ export class FolderUseCases {
   async createFolder(
     creator: User,
     newFolderDto: CreateFolderDto,
+    options?: { transaction?: Transaction },
   ): Promise<Folder> {
     const isAGuestOnSharedWorkspace = creator.email !== creator.bridgeUser;
     let user = creator;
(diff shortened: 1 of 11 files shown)


src/modules/folder/folder.usecase.ts
sg-gs, line 1: Let's keep the transaction just for members joining. Adding it to folders is another topic, and as we aim to strictly ensure that the join is not done multiple times for the same user, this is not required

src/modules/folder/folder.usecase.ts
```diff
@@ -30,6 +30,7 @@ import { WorkspaceAttributes } from '../workspaces/attributes/workspace.attribut
 import { FileUseCases } from '../file/file.usecase';
 import { File, FileStatus } from '../file/file.domain';
 import { CreateFolderDto } from './dto/create-folder.dto';
+import { Transaction } from 'sequelize';
```
sg-gs, line 33: Also, adding database-related things directly here breaks the decoupling the repositories achieve, as this is coupling the use cases to the underlying technology to persist data. Therefore, this is completely discouraged.

src/modules/workspaces/repositories/team.repository.ts
```diff
@@ -135,11 +135,15 @@ export class SequelizeWorkspaceTeamRepository {
...
-    const teamUser = await this.teamUserModel.create({
-      teamId,
-      memberId: userUuid,
-    });
+    const teamUser = await this.teamUserModel.create(
+      {
+        teamId,
+        memberId: userUuid,
+      },
+      { ...options },
```
sg-gs, line 145: By the way, does passing options directly not work?

`.create({...}, options)`

instead of

`.create({...}, {...options})`

src/modules/workspaces/repositories/workspaces.repository.ts
```diff
@@ -84,16 +84,27 @@ export class SequelizeWorkspaceRepository {
...
     where: Partial<WorkspaceInviteAttributes>,
+    options?: { transaction?: Transaction },
   ): Promise<WorkspaceInvite | null> {
-    const invite = await this.modelWorkspaceInvite.findOne({ where });
+    const invite = await this.modelWorkspaceInvite.findOne({
+      where,
+      ...options,
+      lock: options?.transaction
+        ? {
+            level: options.transaction.LOCK.UPDATE,
```
sg-gs, line 103: Nice fine-grained lock 👍🏽

src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -1553,59 +1573,71 @@ export class WorkspacesUsecases {
...
+      const finalMessage = 'There was a problem accepting this invite!';
 
       Logger.error(
         `[WORKSPACE/ACEPT_INVITE]: Error while accepting user invitation! ${
           (error as Error).message
         }`,
       );
 
+      if (error instanceof BadRequestException) {
+        throw new BadRequestException(finalMessage);
```
sg-gs, line 1622: Why are not we returning the real error message in case of a Bad Request?

src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -1553,59 +1573,71 @@ export class WorkspacesUsecases {
...
-      let finalMessage = 'There was a problem accepting this invite! ';
-      const rollbackError = await this.rollbackUserAddedToWorkspace(
-        user.uuid,
-        workspace,
-      );
-
-      finalMessage += rollbackError
-        ? rollbackError.message
-        : 'rollback applied successfully';
+      await transaction.rollback();
```
sg-gs, line 1612: Is there any possibility of setting an expiration for the transaction also, so just in case the connection is lost or everything fails, the database itself expires the lock?

sg-gs (changes requested):

sg-gs:
The coverage requirements are not met also @jzunigax2. You need to increase them a bit.

src/modules/workspaces/workspaces.usecase.ts
jzunigax2: I don't believe it's possible to do so in a sequelize config or inside a sequelize transaction, but it appears it can be configured in the db settings: https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-IDLE-IN-TRANSACTION-SESSION-TIMEOUT

src/modules/workspaces/workspaces.usecase.ts
jzunigax2: Since I moved most of the logic inside a try catch to be able to rollback the transaction our BadRequestExceptions are would be catched and then thrown as 500 Internal server Error, so that's why im rethrowing them as BadRequestException

src/modules/workspaces/workspaces.usecase.ts
sg-gs: Add the flag then

sg-gs (changes requested):
Let's manage the transaction timeout and increase a bit the code coverage and we can go on with this

PR #429, opened 2024-11-11
[_]: feat/remove-encrypted-name-from-search

larryrider:
- Removes the use of encrypted names when searching for files


larryrider:
I have already changed from folder uuid to folder id @sg-gs

sg-gs (approved):

PR #436, opened 2024-11-21
feat: calculate usage incrementally

apsantiso:


migrations/20241120004033-create-calculate-last-day-usage-procedure.js
```diff
@@ -0,0 +1,56 @@
...
+        ) mru
+        ON u.uuid::uuid = mru.user_id::uuid
+        WHERE
+            (mru.last_period IS NOT NULL AND mru.last_period != CURRENT_DATE - INTERVAL '1 day')
+            AND (
+                (f.status != 'DELETED' AND f.created_at BETWEEN CURRENT_DATE - INTERVAL '1 day' AND CURRENT_DATE - INTERVAL '1 millisecond')
+                OR
+                (f.status = 'DELETED' AND f.updated_at BETWEEN CURRENT_DATE - INTERVAL '1 day' AND CURRENT_DATE - INTERVAL '1 millisecond')
+                OR
+                (f.status != 'DELETED' AND f.modification_time BETWEEN CURRENT_DATE - INTERVAL '1 day' AND CURRENT_DATE - INTERVAL '1 millisecond')
```
apsantiso, line 42: I just added this to include files with modified size, but honestly even if we use updated_at, there are some problems with this. I'll explain further later

migrations/20241120004033-create-calculate-last-day-usage-procedure.js
```diff
@@ -0,0 +1,56 @@
...
+            NOW() AS created_at,
+            NOW() AS updated_at
+        FROM
+            files f
+        JOIN users u ON u.id = f.user_id
+        LEFT JOIN (
+            SELECT user_id, MAX(period) AS last_period
+            FROM public.usages
+            GROUP BY user_id
+        ) mru
```
apsantiso, line 33: If a user does not have a record in usage, we would like not to calculate their usage so we can reduce usecases such as an user having their last day usage calculated but files created before this implemented being ignored.

migrations/20241120002050-create-usages-index.js
```diff
@@ -0,0 +1,17 @@
...
+    await queryInterface.sequelize.query(
+      `CREATE INDEX CONCURRENTLY ${indexName} ON usages (user_id, type, period)`,
+    );
+  },
+
+  async down(queryInterface, Sequelize) {
+    await queryInterface.sequelize.query(
+      `DROP INDEX CONCURRENTLY ${indexName}`,
+    );
+  },
```
apsantiso, line 16: This is just a tentative change, we might not need it since it is not really useful for calculating files with changed size

migrations/20241120004033-create-calculate-last-day-usage-procedure.js
sg-gs: I do not get that. If we create the first row on demand for already existing users, which could be the issue?

PR #438, opened 2024-12-02
[_]: feat/switch-from-mailerlite-to-brevo

sg-gs:
We switched Newsletter Mailing providers


sg-gs:
New secrets configured on production: 
- `NEWSLETTER_GROUP_ID`
- `NEWSLETTER_API_KEY`

PR #441, opened 2024-12-13
[PB-3257] Migrate authentication and two-factor endpoints.

evillalba94:
### Ticket
- https://inxt.atlassian.net/browse/PB-3340
This changes are required to work with Access Logs

## Updates
###Endpoints
#### Authentication 
- [POST] `/auth/login`: Return security details
- [POST] `/auth/login/access`: Allow the user access their account
- [GET] `/auth/logout`: This will be used to register log [LOGOUT]

#### Two Factor Authentication
- [GET] `/auth/tfa`: Return code and QR
- [PUT] `/auth/tfa`: Update user 2FA
- [DELETE] `/auth/tfa`: Remove user 2FA


src/modules/auth/auth.controller.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+
+      await authController.loginAccess(loginAccessDto);
+
+      expect(userUseCases.loginAccess).toHaveBeenCalledTimes(1);
+      expect(userUseCases.loginAccess).toHaveBeenCalledWith(loginAccessDto);
+    });
+  });
+
+  describe('GET /logout', () => {
+    it('should return logout response', async () => {
```
apsantiso, line 117: Please use 'When xxx, then yyyy' format to label the tests, we try to make the tests as unified as possible. example:

```
  describe('isBooleanLimit()', () => {
    it('When limit is boolean type, then it should return true', async () => {
      const limit = newFeatureLimit({
        type: LimitTypes.Boolean,
        value: 'false',
      });

      expect(limit.isBooleanLimit()).toBeTruthy();
    });

    it('When limit is not boolean type, then it should return false', async () => {
      const limit = newFeatureLimit({
        type: LimitTypes.Counter,
        value: '3',
      });

      expect(limit.isBooleanLimit()).toBeFalsy();
    });
  });
```

apsantiso (changes requested):

apsantiso:
It might be better to keep TFA endpoints within the `auth` module instead of creating separate modules. Adding a module just for a controller and a single repository (without any use case) can lead to too many modules as the API grows. To keep things manageable, it’s best to avoid adding extra modules unless necessary. Also, the login and loginAccess logic could fit in the auth module. 

However, I understand the register User logic is currently handled in userUsecases. Just my opinion @sg-gs @evillalba94, take it with a grain of salt

src/modules/auth/auth.controller.ts
```diff
@@ -0,0 +1,114 @@
...
+    try {
+      const encryptedSalt = this.cryptoService.encryptText(
+        user.hKey.toString(),
+      );
+      const required2FA = Boolean(
+        user.secret_2FA && user.secret_2FA.length > 0,
+      );
+      const hasKeys = await this.keyServerUseCases.findUserKeys(user.id);
+
+      if (clientId === 'drive-mobile') {
```
sg-gs, line 62: Remove this thing, we no longer offer referrals

src/modules/auth/auth.controller.ts
```diff
@@ -0,0 +1,114 @@
...
+        this.userUseCases
+          .applyReferral(user.id, ReferralKey.InstallMobileApp)
+          .catch((err) => {
+            this.userUseCases.logReferralError(user.id, err);
+          });
+      }
+
+      if (clientId === 'drive-desktop') {
```
sg-gs, line 70: Same with this one

sg-gs (changes requested):

package.json
sg-gs, line 1: The versions of the new libraries are the same as in [drive-server](https://github.com/internxt/drive-server/blob/master/package.json) ✅

src/modules/auth/auth.controller.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+      const result = await authController.login(loginDto, clientId);
+
+      expect(result).toEqual({
+        hasKeys: keys,
+        sKey: 'encryptedText',
+        tfa: true,
+      });
+    });
+
+    it('When user is not found, then it should throw NotFoundException', async () => {
```
sg-gs, line 84: This test is wrong, the current login does not [behave](https://github.com/internxt/drive-server/blob/f6dbe3faab26a68637661083e893607854f66fd8/src/app/routes/auth.ts#L78) like that. On the other hand, we should not inform about if a user exists or not in our service, that allows enumeration attacks to be run.

src/modules/user/user.repository.ts
```diff
@@ -250,6 +251,14 @@ export class SequelizeUserRepository implements UserRepository {
     return this.modelUserNotificationTokens.count({ where: { userId } });
   }
 
+  async loginFailed(uuid: User['uuid'], isFailed: boolean): Promise<void> {
+    const update: any = { errorLoginCount: 0 };
+    if (isFailed) {
+      update.errorLoginCount = Sequelize.literal('error_login_count + 1');
```
sg-gs, line 257: Why? Setting a new number and updating should be enough, this is also business logic (should be in the auth module and the login usecase should call the user's repository to update this field.

src/modules/auth/two-factor-auth.service.ts
```diff
@@ -0,0 +1,42 @@
...
+        qrCode: qrCode,
+      };
+    } catch {
+      throw new InternalServerErrorException(
+        'An error occurred while trying to generate two-factor authentication.',
+      );
+    }
+  }
+
+  async validateTwoFactorAuthCode(secret: string, code: string) {
```
sg-gs, line 30: This does not need to be `async`

src/modules/user/user.usecase.ts
```diff
@@ -1186,4 +1192,135 @@ export class UserUseCases {
...
+    const MAX_LOGIN_FAIL_ATTEMPTS = 10;
+
+    const userData = await this.findByEmail(loginAccessDto.email);
+
+    if (!userData) {
+      Logger.debug(
+        '[AUTH/LOGIN] Attempted login with a non-existing email: %s',
+        loginAccessDto.email,
+      );
+      throw new BadRequestException('Wrong login credentials');
```
sg-gs, line 1206: This [should be](https://github.com/internxt/drive-server/blob/f6dbe3faab26a68637661083e893607854f66fd8/src/app/routes/auth.ts#L128) an Unauthorized. It's important to respect that, as it should work as it currently does; otherwise, we may break multiple things.

src/modules/user/user.usecase.ts
```diff
@@ -1186,4 +1192,135 @@ export class UserUseCases {
...
+      throw new ForbiddenException(
+        'Your account has been blocked for security reasons. Please reach out to us',
+      );
+    }
+
+    const hashedPass = this.cryptoService.decryptText(loginAccessDto.password);
+
+    if (hashedPass !== userData.password.toString()) {
+      await this.loginFailed(userData, true);
+      throw new BadRequestException('Wrong login credentials');
```
sg-gs, line 1222: This [should be](https://github.com/internxt/drive-server/blob/f6dbe3faab26a68637661083e893607854f66fd8/src/app/routes/auth.ts#L141) an Unauthorized. It's important to respect that, as it should work as it currently does; otherwise, we may break multiple things.

src/modules/user/user.usecase.ts
```diff
@@ -1186,4 +1192,135 @@ export class UserUseCases {
...
+    if (twoFactorEnabled) {
+      const tfaResult = speakeasy.totp.verifyDelta({
+        secret: userData.secret_2FA,
+        token: loginAccessDto.tfa,
+        encoding: 'base32',
+        window: 2,
+      });
+
+      if (!tfaResult) {
+        throw new BadRequestException('Wrong 2-factor auth code');
```
sg-gs, line 1236: This [should be](https://github.com/internxt/drive-server/blob/f6dbe3faab26a68637661083e893607854f66fd8/src/app/routes/auth.ts#L154) an Unauthorized. It's important to respect that, as it should work as it currently does; otherwise, we may break multiple things.

sg-gs (changes requested):

PR #442, opened 2024-12-13
[PB-3189] fix: correctly remove children folders from lookup table on delete

apsantiso:
When you delete a folder, its children folders are not being removed correctly from the lookup table, so searching was showing non accesible folders to the users.

This script just modifies 1 line of the function triggered to fix this behavior. For those who want to compare the changes: original migration [here](https://github.com/internxt/drive-server-wip/blob/b3aec38aa0285d701d926895d357a51fe7eb5d62/migrations/20231004145835-add-triggers-for-lookup.js#L122)

**Old condition:**
` ELSIF OLD.deleted = false AND NEW.removed = false AND (NEW.deleted = true OR NEW.removed = true) THEN
`

**New condition**
`ELSIF (OLD.deleted = false OR OLD.removed = false) AND (NEW.deleted = true OR NEW.removed = true) THEN
`

-----
### How to Reproduce the Issue:

1. Move the following folder structure to the trash:

- Parent Folder
  - Child 1
  - Child 2

2. Permanently delete the Parent Folder from the trash.

3. Search for Child 1 or Child 2 using the search bar.

**Expected Result:**

Child 1 and Child 2 should not appear in the search results.

**Actual Result:**

Child 1 and Child 2 still appear in the search results.

### DIFF
--- a/migrations/20241213035931-fix-update-folder-in-lookup-table-function.js
+++ b/migrations/20241213035931-fix-update-folder-in-lookup-table-function.js
@@ -1,0 +1,43 @@
+'use strict';
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.sequelize.query(`
+      CREATE OR REPLACE FUNCTION update_folder_in_look_up_table()
+      RETURNS TRIGGER AS $$
+      DECLARE
+        user_record users%ROWTYPE;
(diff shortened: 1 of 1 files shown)


sg-gs:
Migration run @apsantiso

sg-gs (approved):

PR #444, opened 2024-12-16
[PB-3334] Feat/access logs

evillalba94:
### Ticket
- https://inxt.atlassian.net/browse/PB-3334

#### Endpoint
- GET: `workspaces/:workspaceId/access/logs` (`ONLY FOR OWNER`)

### AccessLogs: Database, Model, Domain, Attributes
- New Table: workspace_logs

### Interceptor
- Contains the logic for create the logs. We must set the endpoints from which we want to log along with their log type using a new decorator. Example: `@WorkspaceLogAction(WorkspaceLogType.SHARE)`. 

## IMPORTANT NOTE
- The interceptor decorator must be integrated on the authentication endpoints.

### Test Endpoint

image

image


src/modules/workspaces/attributes/workspace-logs.attributes.ts
```diff
@@ -0,0 +1,34 @@
+export enum WorkspaceLogType {
+  LOGIN = 'LOGIN',
```
sg-gs, line 2: The [standard](https://www.typescriptlang.org/docs/handbook/enums.html) is to use camelCase for the enum keys and you can also use lowercase for the values like: 

```
export enum WorkspaceLogType {
  LogIn = 'login',
  PasswordUpdated = 'password-updated',
  ...
}
```

src/modules/workspaces/attributes/workspace-logs.attributes.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export interface WorkspaceLogAttributes {
+  id: string;
+  workspaceId: string;
+  creator: string;
+  type: WorkspaceLogType;
+  platform: WorkspaceLogPlatform;
+  entityId?: string;
+  user?: any;
```
sg-gs, line 28: Do not use `any`

src/modules/workspaces/attributes/workspace-logs.attributes.ts
```diff
@@ -0,0 +1,34 @@
...
+  workspace?: any;
+  file?: any;
```
sg-gs, line 30: Same here

src/modules/workspaces/attributes/workspace-logs.attributes.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export interface WorkspaceLogAttributes {
+  id: string;
+  workspaceId: string;
+  creator: string;
+  type: WorkspaceLogType;
+  platform: WorkspaceLogPlatform;
+  entityId?: string;
+  user?: any;
+  workspace?: any;
```
sg-gs, line 29: And here

src/modules/workspaces/attributes/workspace-logs.attributes.ts
```diff
@@ -0,0 +1,34 @@
...
+  file?: any;
+  folder?: any;
```
sg-gs, line 31: And here

src/modules/workspaces/domains/workspace-log.domain.ts
sg-gs, line 1: Do not use `any`

src/modules/workspaces/guards/workspaces.guard.ts
```diff
@@ -57,13 +64,22 @@ export class WorkspaceGuard implements CanActivate {
...
+
     if (accessContext === AccessContext.WORKSPACE) {
-      return this.verifyWorkspaceAccessByRole(user, id, requiredRole);
+      verified = await this.verifyWorkspaceAccessByRole(user, id, requiredRole);
     } else if (accessContext === AccessContext.TEAM) {
-      return this.verifyTeamAccessByRole(user, id, requiredRole);
+      verified = await this.verifyTeamAccessByRole(user, id, requiredRole);
+    }
+
+    if (verified && workspaceLogAction === WorkspaceLogType.DELETE_ALL) {
```
sg-gs, line 77: I do not understand DELETE_ALL usage, neither the getItems request. There is no DELETE_ALL log whatsoever contemplated on the product request

src/modules/workspaces/interceptors/workspaces-logs.interceptor.ts
```diff
@@ -0,0 +1,332 @@
...
+    };
+  }
+
+  determinePlatform(client: string): WorkspaceLogPlatform {
+    const platforms = {
+      'drive-web': WorkspaceLogPlatform.WEB,
+      'drive-mobile': WorkspaceLogPlatform.MOBILE,
+      'drive-desktop': WorkspaceLogPlatform.DESKTOP,
+    };
+    return platforms[client] || WorkspaceLogPlatform.UNSPECIFIED;
```
sg-gs, line 56: The platform should be always specified, if not, the log does not have sense

sg-gs (changes requested):

src/modules/workspaces/interceptors/workspaces-logs.interceptor.ts
```diff
@@ -0,0 +1,332 @@
...
+
+    if (!action) {
+      Logger.debug(
+        `[WORKSPACE/LOGS] Invalid action type: ${type} or item type: ${itemType}`,
+      );
+      return null;
+    }
+
+    return action;
+  }
```
apsantiso, line 309: Honestly, this just adds complexity for such a simple function you just used twice. It kinda makes hard to understand what you're trying to accomplish. 

I know you want to encapsulate things, but sometimes jumping between code, specially something just called `determineAction`, forces other devs to try to understand what you are trying to do

src/modules/workspaces/interceptors/workspaces-logs.interceptor.ts
```diff
@@ -0,0 +1,332 @@
...
+              creator: requesterUuid,
+              type: action,
+              platform,
+              entityId: item.uuid || item.id,
+            });
+          }
+          return null;
+        })
+        .filter((promise): promise is Promise<any> => promise !== null);
+      await Promise.all(deletePromises);
```
apsantiso, line 156: This function also applies when I empty my trash, right?  (I am just guessing as there is a DELETE_ALL action)

What happens if user has 1000 items inside the trash? Are we going to send 1000 inserts to the database at the same time? This is something we could discuss with business

apsantiso (changes requested):

src/modules/workspaces/interceptors/workspaces-logs.interceptor.ts
```diff
@@ -0,0 +1,332 @@
...
+      workspaceIds.map((workspaceId) =>
+        this.registerWorkspaceLog({
+          workspaceId,
+          creator: user.uuid,
+          type: actionType,
+          platform,
+        }),
+      ),
+    );
+  }
```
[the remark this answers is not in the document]
evillalba94: It was a business decision. @sg-gs 
We are making the log in as a global aspect, just when the user log in the application.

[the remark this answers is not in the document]
apsantiso: Ok, I now see you are just using this action for actions that involve actions tied to the user account. Please use a less generic name so people do not try to use this for other actions

src/modules/workspaces/interceptors/workspaces-logs.interceptor.ts
evillalba94: @apsantiso This was thinking about the future, since you may want to record other logs such as `move` or `rename` which are applied to files and folders. However, for the moment it was not considered for this first approach. In case later on we need to add both here we would not need to create so many if for validation.

Anyway, let me know if I destroy the function and add the number of if separately.

apsantiso: If you find too much trouble while trying to remove the function, then try to change its name to something more specific, like `determineItemAction`, so we can know this is something used specifically to determine what action is being used on an item.

src/modules/workspaces/interceptors/workspaces-logs.interceptor.spec.ts
```diff
@@ -0,0 +1,699 @@
...
+      const platform = interceptor.determinePlatform('drive-mobile');
+      expect(platform).toBe(WorkspaceLogPlatform.Mobile);
+    });
+
+    it('When client is drive-desktop, then it should return DESKTOP platform', () => {
+      const platform = interceptor.determinePlatform('drive-desktop');
+      expect(platform).toBe(WorkspaceLogPlatform.Desktop);
+    });
+
+    it('When client is unknown, then it should return UNSPECIFIED platform', () => {
```
sg-gs, line 40: I assume this test should no longer be needed, as we are not allowing unknown platforms

sg-gs (changes requested):

apsantiso (approved):

sg-gs (approved):

sg-gs:
Migrations run @evillalba94 @apsantiso 

```

== 20241113173746-create-workspace-logs-table: migrating =======
== 20241113173746-create-workspace-logs-table: migrated (0.656s)

✨  Done in 2.96s.

```

PR #448, opened 2025-01-06
[PB-3595] feat: migrated credentials and email verification endpoints

apsantiso:
This migrates the are-credentials-correct and verification email related endpoints to this API.

We need to add the next variable to the environment before merging:

`SENDGRID_TEMPLATE_DRIVE_EMAIL_VERIFICATION`

**The daily mail limit reached error now throws a valid 429 response instead of a 500 error like the old API.**


src/modules/user/user.usecase.spec.ts
sg-gs, line 1: Add a test to check that the mail limit increase method (`increaseTodayAttemps`) increases the number of the attempts. Otherwise, a failure there will make the daily limit useless.

sg-gs (approved):

PR #449, opened 2025-01-10
[_] chore: modified verification endpoints to a more restful name

apsantiso:


src/modules/user/user.controller.ts
```diff
@@ -497,7 +497,9 @@ export class UserController {
     @Res({ passthrough: true }) res: Response,
   ) {
     try {
-      await this.userUseCases.sendAccountRecoveryEmail(body.email.toLowerCase());
+      await this.userUseCases.sendAccountRecoveryEmail(
+        body.email.toLowerCase(),
+      );
```
apsantiso, line 502: Just a linter change, nothing modified here.

PR #455, opened 2025-01-14
[PB-3601] feat: migrate user endpoints

evillalba94:
### Ticket
- https://inxt.atlassian.net/browse/PB-3601

### Endpoints
- `PUT /api/users/avatar`
- `DELETE /api/users/avatar `
- `PATCH /api/users/profile`

## Note about a Bug
image
The error that persisted in the tests of `user.controller.spec.ts` that did not allow this task to finish was due to the order of the imports. For some reason, when adding the code related to loading the user's avatar to the controller, the test file (`user.controller.spec.ts`) stopped working. After doing several investigations and changes, it was discovered that moving the imports could solve the problem (the import of `generateBase64PrivateKeyStub` was moved before the configuration one). I leave this here for future similar errors.

### DIFF
--- a/src/modules/user/dto/update-profile.dto.ts
+++ b/src/modules/user/dto/update-profile.dto.ts
@@ -1,19 +1,48 @@
 import { ApiProperty } from '@nestjs/swagger';
-import { IsNotEmpty } from 'class-validator';
+import {
+  IsNotEmpty,
+  IsOptional,
+  IsString,
+  MaxLength,
+  ValidateIf,
+} from 'class-validator';
 import { UserAttributes } from '../user.attributes';
 
 export class UpdateProfileDto {
-  @IsNotEmpty()
+  @IsOptional()
+  @IsString()
+  @ValidateIf((o) => o.name !== null)
+  @IsNotEmpty({ message: 'Name should not be empty if provided.' })
+  @MaxLength(100, { message: 'Name must be at most 100 characters long.' })
   @ApiProperty({
     example: 'Internxt',
     description: 'Name of the new user',
   })
-  name: UserAttributes['name'];
+  name?: UserAttributes['name'];
 
-  @IsNotEmpty()
+  @ValidateIf((o) => o.name === null)
+  @IsNotEmpty({ message: 'Name should not be null if provided.' })
+  @ApiProperty({
+    example: 'Internxt',
+    description: 'Name of the new user',
+  })
+  nameNullCheck?: string;
+
+  @IsOptional()
+  @IsString()
+  @ValidateIf((o) => o.lastname !== null)
+  @MaxLength(100, { message: 'Lastname must be at most 100 characters long.' })
   @ApiProperty({
     example: 'Lastname',
     description: 'Last name of the new user',
   })
-  lastname: UserAttributes['lastname'];
+  lastname?: UserAttributes['lastname'] | '';
+
+  @ValidateIf((o) => o.lastname === null)
+  @IsNotEmpty({ message: 'Lastname should not be null if provided.' })
+  @ApiProperty({
+    example: 'Lastname',
+    description: 'Last name of the new user',
+  })
+  lastnameNullCheck?: string;
 }
--- a/src/modules/user/dto/update-profile.dto.ts
+++ b/src/modules/user/dto/update-profile.dto.ts
@@ -1,0 +1,19 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { IsNotEmpty } from 'class-validator';
+import { UserAttributes } from '../user.attributes';
+
+export class UpdateProfileDto {
+  @IsNotEmpty()
+  @ApiProperty({
+    example: 'Internxt',
+    description: 'Name of the new user',
+  })
+  name: UserAttributes['name'];
+
+  @IsNotEmpty()
+  @ApiProperty({
+    example: 'Lastname',
+    description: 'Last name of the new user',
+  })
+  lastname: UserAttributes['lastname'];
+}
--- a/src/modules/user/dto/update-profile.dto.ts
+++ b/src/modules/user/dto/update-profile.dto.ts
@@ -36,7 +36,7 @@ export class UpdateProfileDto {
     example: 'Lastname',
     description: 'Last name of the new user',
   })
-  lastname?: UserAttributes['lastname'] | '';
+  lastname?: UserAttributes['lastname'];
 
   @ValidateIf((o) => o.lastname === null)
   @IsNotEmpty({ message: 'Lastname should not be null if provided.' })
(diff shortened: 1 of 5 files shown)


src/modules/user/user.controller.spec.ts
```diff
@@ -253,4 +263,70 @@ describe('User Controller', () => {
...
     });
   });
+
+  describe('PATCH /profile', () => {
+    it('When updateProfile is called then it should update the user profile', async () => {
+      const user = newUser();
+      const payload: UpdateProfileDto = { name: 'John', lastname: 'Doe' };
+
+      await userController.updateProfile(user, payload);
+      expect(userUseCases.updateProfile).toHaveBeenCalledWith(user, payload);
```
sg-gs, line 273: What if the name or the last name is longer than the database field? how are we controlling that case?

And what happens if `name` or `lastname` is missing?

src/modules/user/user.usecase.ts
```diff
@@ -1377,6 +1379,65 @@ export class UserUseCases {
     );
   }
 
+  async upsertAvatar(
+    user: User,
+    avatarKey: string,
+  ): Promise<Error | { avatar: string }> {
+    if (!avatarKey) {
+      throw new BadRequestException('Avatar key required');
```
sg-gs, line 1387: This should be on the controller. The type is not suggesting that it is optional and thus, undefined or null - nor it should do -. It's the controller's responsability to validate this thing.

src/modules/user/user.usecase.ts
```diff
@@ -1377,6 +1379,65 @@ export class UserUseCases {
...
+    }
+
+    if (user.avatar) {
+      try {
+        await this.avatarService.deleteAvatar(user.avatar);
+      } catch (err) {
+        Logger.error(
```
sg-gs, line 1394: This log should be on the controller

src/modules/user/user.usecase.ts
```diff
@@ -1377,6 +1379,65 @@ export class UserUseCases {
...
+    }
+
+    try {
+      await this.userRepository.updateById(user.id, {
+        avatar: avatarKey,
+      });
+      const avatarUrl = await this.getAvatarUrl(avatarKey);
+      return { avatar: avatarUrl };
+    } catch (err) {
+      Logger.error(
```
sg-gs, line 1410: This log should be on the controller.

src/modules/user/user.usecase.ts
```diff
@@ -1377,6 +1379,65 @@ export class UserUseCases {
...
+  async deleteAvatar(user: User): Promise<Error | void> {
+    if (user.avatar) {
+      try {
+        await this.avatarService.deleteAvatar(user.avatar);
+        await this.userRepository.updateById(user.id, {
+          avatar: null,
+        });
+      } catch (err) {
+        Logger.error(
+          `[USER/DELETE_AVATAR]Deleting the avatar for the user: ${
```
sg-gs, line 1428: This log should be on the controller.

sg-gs (changes requested):

src/modules/user/user.controller.spec.ts
evillalba94: @sg-gs 
The update-profile DTO and controller were changed to validate the discussed aspects.

- `name` cannot be `null` or `''`.
- `lastname` cannot be `null` but can be `''`.
- `name` or `lastname` cannot be longer than 100 characters.
- At least `name` or `lastname` must come in the request.
- If `name` or `lastname` is received in the request, the update will be executed.

src/modules/user/dto/update-profile.dto.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  @IsOptional()
+  @IsString()
+  @ValidateIf((o) => o.lastname !== null)
+  @MaxLength(100, { message: 'Lastname must be at most 100 characters long.' })
+  @ApiProperty({
+    example: 'Lastname',
+    description: 'Last name of the new user',
+  })
+  lastname?: UserAttributes['lastname'] | '';
```
sg-gs, line 39: The thing was to just make the `lastname` optional, I do not understand why the other field is needed also.

src/modules/user/user.controller.ts
```diff
@@ -835,4 +841,72 @@ export class UserController {
...
+    summary: 'Update user profile',
+  })
+  @ApiOkResponse({
+    description: 'Updated user profile',
+  })
+  async updateProfile(
+    @UserDecorator() user: User,
+    @Body() updateProfileDto: UpdateProfileDto,
+  ) {
+    if (!updateProfileDto.name && updateProfileDto.lastname == undefined) {
```
sg-gs, line 858: Making the `lastname` optional and the `name` not makes this check unnecesary. As long as the requester sends `name` or `name` and `lastname`, everything should be fine

src/modules/user/user.usecase.ts
```diff
@@ -1377,6 +1379,46 @@ export class UserUseCases {
...
 
+  async upsertAvatar(
+    user: User,
+    avatarKey: string,
+  ): Promise<Error | { avatar: string }> {
+    if (user.avatar) {
+      try {
+        await this.avatarService.deleteAvatar(user.avatar);
+      } catch (err) {
+        throw new InternalServerErrorException('Failed to delete old avatar');
```
sg-gs, line 1390: We are losing the original error here. The trick would be to keep the message, the stack, etc while throwing the InternalServerErrorException to prevent the error to be leaked to the requester.

src/modules/user/user.usecase.ts
```diff
@@ -1377,6 +1379,46 @@ export class UserUseCases {
...
+    }
+
+    try {
+      await this.userRepository.updateById(user.id, {
+        avatar: avatarKey,
+      });
+      const avatarUrl = await this.getAvatarUrl(avatarKey);
+      return { avatar: avatarUrl };
+    } catch (err) {
+      throw new InternalServerErrorException('Failed to add new avatar');
```
sg-gs, line 1401: same here

src/modules/user/user.usecase.ts
```diff
@@ -1377,6 +1379,46 @@ export class UserUseCases {
...
+
+  async deleteAvatar(user: User): Promise<Error | void> {
+    if (user.avatar) {
+      try {
+        await this.avatarService.deleteAvatar(user.avatar);
+        await this.userRepository.updateById(user.id, {
+          avatar: null,
+        });
+      } catch (err) {
+        throw new InternalServerErrorException('Failed to delete avatar');
```
sg-gs, line 1413: same here

sg-gs (changes requested):

apsantiso (approved):

sg-gs (approved):

PR #458, opened 2025-01-21
[_] feat: add support for multiple keys update password, login, user creation

apsantiso:
This reverts commit adcd1d014c5cdc51fc24159151333cce624a9f7f, reversing changes made to aa9f82d97dc3aa013667c24bd10d8c72986056b6.

Adds again the multi key support. 

New changes besides the revert:

1. Kyber keys object enforced validation has been removed temporarily
2. Keys nested validation is not enforced temporarily
3. Added temporarily validation on the controllers 
4. Update password now fails if user has kyber keys and the client does not send any kyber key.
5. Increased public_key and private_key fields length on the database

TODO comments mark the parts we need to reverse when the frontend aligns with the backend expected object.

For extra information you can check https://github.com/internxt/drive-server-wip/pull/450, which is the PR we intend to merge again.

### DIFF
--- a/src/modules/user/user.controller.spec.ts
+++ b/src/modules/user/user.controller.spec.ts
@@ -1,4 +1,5 @@
 import { DeepMocked, createMock } from '@golevelup/ts-jest';
+import { v4 } from 'uuid';
 import {
   BadRequestException,
   ForbiddenException,
@@ -7,7 +8,11 @@ import {
   UnauthorizedException,
 } from '@nestjs/common';
 import * as jwtUtils from '../../lib/jwt';
-import { newUser } from '../../../test/fixtures';
+import {
+  newKeyServer,
+  newPreCreatedUser,
+  newUser,
+} from '../../../test/fixtures';
 import getEnv from '../../config/configuration';
 import { UserController } from './user.controller';
 import { MailLimitReachedException, UserUseCases } from './user.usecase';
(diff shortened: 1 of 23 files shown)


src/modules/user/user.controller.spec.ts
```diff
@@ -45,6 +54,14 @@ jest.mock('../../config/configuration', () => {
...
 });
 
+jest.mock('../../lib/jwt', () => {
+  return {
+    ...jest.requireActual('../../lib/jitsi'),
+    generateJitsiJWT: jest.fn(() => 'newJitsiJwt'),
+    verifyWithDefaultSecret: jest.fn(() => 'defaultVerifiedSecret'),
+  };
+});
+
```
apsantiso, line 64: Added this mock as calling the real generateBase64PrivateKeyStub() just makes this test too slow and sometimes jest just exited gracefully due to timeout. 

Test cases for jitsi's endpoints do not require it.

src/modules/user/user.controller.spec.ts
```diff
@@ -149,6 +166,14 @@ describe('User Controller', () => {
...
     it('When token and user are correct, then resolves', async () => {
       userUseCases.unblockAccount.mockResolvedValueOnce();
+      jest.spyOn(jwtUtils, 'verifyWithDefaultSecret').mockReturnValueOnce({
+        payload: {
+          uuid: user.uuid,
+          email: user.email,
+          action: AccountTokenAction.Unblock,
+        },
+        iat: 123123,
+      });
```
apsantiso, line 176: This test for `/account-unblock` requires JWT library returned value.

src/modules/user/user.controller.spec.ts
apsantiso: Handled here https://github.com/internxt/drive-server-wip/pull/460

src/modules/user/user.controller.spec.ts
apsantiso: handled here https://github.com/internxt/drive-server-wip/pull/460

apsantiso:
This PR vas merged in two parts.

PR #461, opened 2025-01-28
[PB-3625] feat: add kyber keys support to update password

apsantiso:
### Changes
- Added missing tests to affected endpoints to ensure backwards compatibility (we just want to check privateKey, publicKey and revocationKey are being returned as expected)

- **GET /public-key/:email**:
  - Continues sending `publicKey` field for backward compatibility.
  - new keys object containing kyber and ecc keys is returned.

- **Put /password/**:
  - Added privateKyberKey as optional body input.
  - If users have Kyber keys but a new encrypted Kyber key is not provided, the password update is prevented.
  - Removed the private key creation function from the logic. This function was a potential source of issues; in rare cases (e.g., when users without any keys tried to update their password), it could result in empty rows with only privateKeys being created.

- **Post /users/**:
  - Keys are currently created as ECC keys. Kyber support will be added to this endpoint in another PR.

- **POST /pre-created-users/register**:
  - Keys are currently created as ECC keys. Kyber support will be added to this endpoint in another PR.

### DIFF
--- a/src/modules/keyserver/key-server.usecase.ts
+++ b/src/modules/keyserver/key-server.usecase.ts
@@ -1,46 +1,137 @@
-import { Injectable } from '@nestjs/common';
+import { Injectable, Logger } from '@nestjs/common';
 import { UserAttributes } from '../user/user.attributes';
-import { Keys } from './key-server.domain';
+import {
+  KeyServer,
+  KeyServerAttributes,
+  UserKeysEncryptVersions,
+} from './key-server.domain';
 import { SequelizeKeyServerRepository } from './key-server.repository';
+
+export class InvalidKeyServerException extends Error {
+  constructor(public validationMessage: string) {
+    super(validationMessage);
+    Object.setPrototypeOf(this, InvalidKeyServerException.prototype);
+  }
+}
+
+type PartialKeys = Partial<Omit<KeyServerAttributes, 'id' | 'userId'>>;
 
 @Injectable()
 export class KeyServerUseCases {
   constructor(private repository: SequelizeKeyServerRepository) {}
 
-  async addKeysToUser(userId: UserAttributes['id'], keys: Keys): Promise<Keys> {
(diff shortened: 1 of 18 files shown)


src/modules/user/user.usecase.spec.ts
```diff
@@ -1028,6 +1033,48 @@ describe('User use cases', () => {
       expect(result).toHaveProperty('token', 'authToken');
       expect(result).toHaveProperty('newToken', 'newAuthToken');
     });
+
+    it('When user without keys logs and no key is saved, then it should return empty keys ', async () => {
```
sg-gs, line 1037: _When user without keys logs in_

src/modules/user/user.controller.spec.ts
```diff
@@ -443,4 +452,315 @@ describe('User Controller', () => {
...
+
+      const result = await userController.updatePassword(
+        mockUpdatePasswordDto,
+        mockUser,
+        clientId,
+      );
+
+      expect(userUseCases.updatePassword).toHaveBeenCalledWith(mockUser, {
+        currentPassword: mockUser.password,
+        newPassword: 'decryptedNewPassword',
```
sg-gs, line 574: Extract this on a variable, same with the salt

src/modules/keyserver/key-server.usecase.ts
```diff
@@ -1,46 +1,137 @@
...
 import { UserAttributes } from '../user/user.attributes';
-import { Keys } from './key-server.domain';
+import {
+  KeyServer,
+  KeyServerAttributes,
+  UserKeysEncryptVersions,
+} from './key-server.domain';
 import { SequelizeKeyServerRepository } from './key-server.repository';
 
+export class InvalidKeyServerException extends Error {
```
sg-gs, line 10: Extend from a BadRequestException instead of from a generic Error, we coupled the code to nestjs as it itself abstracts these things.

sg-gs (changes requested):

sg-gs (approved):

PR #464, opened 2025-01-29
[PB-3625] feat: added support for kyber keys on log in and user creation

apsantiso:
### Updated Endpoints

- **POST /login**:
  - Returns `hasKyberKeys` and `hasEccKeys` flags instead of a single `hasKeys` flag.
  - Allows differentiation between the presence of Kyber and ECC keys.

- **POST /login/access**:
  - Accepts keys via the `keys` object (`KeysDto`) or the legacy fields for ECC encryption (for backwards compatibility)
  - If both are provided, the keys inside the `keys` object take priority.
  - Returns ECC keys in the legacy structure and includes the `keys` object containing both ECC and Kyber keys.

- **POST /users**:
  - Now supports receiving keys via the `keys` object (`KeysDto`).
  - Enables submission of both ECC and Kyber keys during user creation.
  
- **POST /pre-created-users/register**:
  - Now supports receiving keys via the `keys` object (`KeysDto`).
  - Allows submission of both ECC and Kyber keys during the registration of pre-created users.

### DIFF
--- a/src/modules/keyserver/key-server.usecase.ts
+++ b/src/modules/keyserver/key-server.usecase.ts
@@ -6,6 +6,7 @@ import {
   UserKeysEncryptVersions,
 } from './key-server.domain';
 import { SequelizeKeyServerRepository } from './key-server.repository';
+import { EccKeysDto, KyberKeysDto } from './dto/keys.dto';
 
 export class InvalidKeyServerException extends BadRequestException {
   constructor(public validationMessage: string) {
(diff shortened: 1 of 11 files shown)


src/modules/keyserver/key-server.usecase.ts
```diff
@@ -134,4 +135,38 @@ export class KeyServerUseCases {
...
+      privateKey: string;
+      publicKey: string;
+      revocationKey: string;
+    },
+  ): {
+    ecc: EccKeysDto;
+    kyber: KyberKeysDto;
+  } {
+    const eccKeys =
+      keys?.ecc || (oldKeys?.publicKey && oldKeys?.privateKey)
```
apsantiso, line 154: Just a helper not to repeat code as we could still receive the keys via the old fields if needed.

sg-gs (approved):

PR #465, opened 2025-01-31
[_] feat: add hybrid encryption support for pre created users

apsantiso:
We need to add support for pre-created users' Kyber keys. This PR enables the backend to generate temporary Kyber keys while also allowing hybrid-encrypted invitations to be updated with frontend-generated keys when the user registers.

### **Changes**  

- Added support for pre-created users' Kyber keys.  
- The backend now generates temporary Kyber keys for pre-created users.  
- Hybrid-encrypted invitations are updated with frontend-generated keys when the user registers.  
- Hybrid-encrypted invitations that belongs to **pre-created users** without any kyber key and/or new kyber keys are not re generated when users register, get deleted. This is an edge case but it is possible if user registers from any source that does not provide kyber keys.
- Added an import helper for ESM packages. Since NestJS compiles into CommonJS and is written in TypeScript, we cannot import ESM libraries traditionally. NestJS does not support ESM yet, and a standard dynamic import won't work due to TypeScript compilation. See [this example in NestJS's official repo](https://github.com/nestjs/nest/tree/master/sample/34-using-esm-packages).  
- Added Kyber KEM as a provider, initializing it on bootstrap to avoid using dynamic imports on every request.  

### **Updated Endpoints**  

- **`POST /users/`**: Now updates hybrid-encrypted invitations when a pre-created user registers.  
- **`POST /users/pre-created/register`**: Now updates hybrid-encrypted invitations when a pre-created user registers.  
- **`POST /users/pre-created/`**: Now generates Kyber keys alongside ECC keys.


sg-gs (commented):
I find the cryptography tests a bit weak. The main reason is that we are mocking the underlying libraries doing the hard thing, which is not the best way to ensure this does not get broken. 

It would be better to use NIST cases hardcoded here so we assert the same things that some of the best cryptography experts in the world consider valid to ensure the validity of algorithms like Kyber. 

Therefore, I invite @TamaraFinogina here to join the review as she has done that on drive-web

src/externals/asymmetric-encryption/utils.spec.ts
```diff
@@ -0,0 +1,73 @@
+import { extendSecret, XORhex } from './utils';
+
+describe('extendSecret', () => {
+  it('When given a secret, then it should return a hashed value with the desired bit-length', async () => {
```
TamaraFinogina, line 4: We must check that the front and back extend the secret identically, or decryption will fail. Since the extend secret is blake3 hash, we can test it using values from their official repo: https://github.com/BLAKE3-team/BLAKE3/blob/master/test_vectors/test_vectors.json

as it is done drive-web:
https://github.com/internxt/drive-web/blob/master/test/unit/services/utils.hash.test.ts

src/externals/asymmetric-encryption/asymmetric-encryption.service.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+        publicKeyInBase64: 'ECCPublicKey',
+      });
+
+      expect(result).toContain(expectedBase64);
+      expect(result).not.toContain('$');
+    });
+  });
+
+  describe('hybridDecryptMessageWithPrivateKey', () => {
+    it('When both ECC and Kyber keys are passed, then it should decrypt using hybrid mode', async () => {
```
TamaraFinogina, line 140: same here - should be a real test

src/externals/asymmetric-encryption/asymmetric-encryption.service.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+
+      expect(keys).toEqual({
+        publicKey: Buffer.from(publicKey).toString('base64'),
+        privateKey: Buffer.from(privateKey).toString('base64'),
+      });
+    });
+  });
+
+  describe('encapsulateWithKyber', () => {
+    it('When called, then it should encapsulate secret with kyber keys', async () => {
```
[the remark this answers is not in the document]
TamaraFinogina: @sg-gs The Kyber library we use ([Dashlane](https://github.com/Dashlane)) was released 3 years ago (before NIST's final publication). Unfortunately, NIST made small changes before standardizing Kyber (basically removed a couple of hashes that were done during encapsulation/decapsulation). It's nothing major, but still slightly different from the version we use. So, we cannot use the latest test vectors.

Btw, since Dashlane is a wrapper for [PQClean](https://github.com/PQClean/PQClean) library, which is up to date with all NIST changes, I've asked the Dashlane author to re-release new pre-build bindings.

apsantiso:
I will take care of this once I clear the priority tasks I have assigned right now. I need to check how viable is testing the real library as we are importing it with a bit of "magic" due to Nestjs limitation of compiling into CommonJS. @TamaraFinogina @sg-gs

package.json
```diff
@@ -13,9 +13,9 @@
     "start:debug": "cross-env NODE_ENV=development nest start --debug --watch",
     "start:prod": "cross-env NODE_ENV=production node -r newrelic dist/main",
     "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
-    "test": "cross-env NODE_ENV=test jest",
-    "test:watch": "jest --watch",
-    "test:cov": "jest --coverage",
+    "test": "NODE_ENV=test NODE_OPTIONS=--experimental-vm-modules jest",
+    "test:watch": "NODE_OPTIONS=--experimental-vm-modules jest --watch",
+    "test:cov": "NODE_OPTIONS=--experimental-vm-modules jest --coverage",
```
apsantiso, line 18: Required to use ESM modules (compiled to commonjs) with Jest.

apsantiso:
I added tests using the real library and making sure original message is the same after we re-encrypt the sharing invitations. This is ready for review again

@sg-gs @TamaraFinogina

src/modules/user/user.usecase.ts
```diff
@@ -466,35 +470,54 @@ export class UserUseCases {
...
-        privateKeyInBase64: privateKey,
-      });
+      const { encryptionKey } = invite;
 
-      const newEncryptedEncryptionKey = await encryptMessageWithPublicKey({
-        message: decryptedEncryptionKey.toString(),
-        publicKeyInBase64: newPublicKey,
-      });
+      if (invite.isHybrid() && (!newPublicKyberKey || !privateKyberKey)) {
+        await this.sharingRepository.deleteInvite(invite);
```
TamaraFinogina, line 491: Can this happen?

TamaraFinogina (approved):

apsantiso:
@sg-gs @TamaraFinogina I will split this PR in a way it does not break our flows, just want to be sure how to merge this in a safer way! I will add the new encryption service first along with the tests @TamaraFinogina already approved so we can merge migration-business logic later!

apsantiso:
Closing this PR as I already separated this into multiple PRs.

https://github.com/internxt/drive-server-wip/pull/493
https://github.com/internxt/drive-server-wip/pull/491
https://github.com/internxt/drive-server-wip/pull/492

PR #481, opened 2025-03-10
[PB-4051] feat: migrated usage endpoint and implement cache module

apsantiso:
We need to continue with the migration of the endpoints in drive-server to this new API. This PR intends to migrate the /usage endpoint while also implementing caching using Nestjs Cache manager. 

### Changes

1. Migrated users/storage/usage endpoint.
2. Implemented Cache manager.

### Requirements

1. Add `REDIS_CONNECTION_STRING` to drive-server-wip if still not set.


sg-gs:
The `REDIS_CONNECTION_STRING` variable is already set. Although you have conflicts with master @apsantiso

src/modules/user/user.controller.ts
```diff
@@ -969,4 +970,16 @@ export class UserController {
 
     return this.userUseCases.confirmDeactivation(token);
   }
+
+  @Get('/storage/usage')
```
sg-gs, line 974: Just use /usage if it is not an issue, as it could be more semantic, to get the usage a user is doing of Drive, which now implies the bytes, but it could imply more things in the future

sg-gs (approved):

src/modules/user/user.controller.ts
apsantiso: Done

PR #484, opened 2025-03-13
[PB-3705] feat: migrate storage endpoints

evillalba94:
### Updates
- Migrating storage-related endpoints and database deletions
---

### Endpoints

- `drive-server` `DELETE` _/storage/folder/:id_ `->` `wip`  _/storage/trash/folder/:folderId_
Param: `:folderId` == Folder.id
- `drive-server` `DELETE` _/storage/folder/:folderid/file/:fileid_ 
`drive-server` `DELETE` /storage/bucket/:bucketid/file/:fileid_  `->` `wip` _/files/:uuid_
Param: `:uuid` == File.uuid


apsantiso (changes requested):

src/modules/trash/trash.controller.ts
```diff
@@ -335,15 +335,23 @@ export class TrashController {
...
 
-    await this.trashUseCases.deleteItems(user, [], [folders[0]]);
+    await this.folderUseCases.deleteByUser(user, [folder]);
+
+    this.storageNotificationService.folderDeleted({
+      payload: { id: folderId, userId: user.id },
+      user: user,
+      clientId,
+    });
   }
```
apsantiso, line 356: El endpoint para borrar folders es `trash/folders/:folderId` y el de files `files/:uuid`. Podrias crear un endpoint `DELETE folders/:uuid` en el controller de folders para mantener la misma estructura de los endpoints? 

Y por fa, en ese endpoint usa UUID en vez del ID. Este metodo lo puedes dejar como estaba antes si quieres

apsantiso (changes requested):
Solo cambia lo del endpoint para tratar de tener los endpoints lo mas estructurados posibles y ya puedes pedirle a @sg-gs que corra los scripts

apsantiso (dismissed):

sg-gs:
This PR does multiple things and is well above >500 LOCs @evillalba94. Hence, you need to split it and try to make a PR per thing. I suggest splitting it into at least two: the trigger update and the migrated endpoints. If any of them is above 500LOCs, divide it again (per endpoint, for instance) @evillalba94

src/modules/file/file.usecase.ts
```diff
@@ -75,8 +79,35 @@ export class FileUseCases {
...
+
+    const { id, uuid, fileId, bucket } = file;
+    await this.network.deleteFile(user, bucket, fileId);
+    await this.shareUsecases.deleteFileShare(id, user);
+    await this.sharingUsecases.bulkRemoveSharings(
+      user,
+      [uuid],
+      SharingItemType.File,
+    );
+    await this.thumbnailUsecases.deleteBy({ fileId: id });
```
apsantiso, line 104: On drive-server we delete the thumbnails files from the buckets and the database, don't you need to do the same here?

src/modules/file/file.usecase.ts
```diff
@@ -75,8 +79,35 @@ export class FileUseCases {
...
+    await this.fileRepository.destroyFile({
+      userId: user.id,
+      fileId,
+      bucket,
+    });
```
apsantiso, line 109: @sg-gs  Should we mimic [drive-server's approach](https://github.com/internxt/drive-server/blob/c58a8912441460e2dce444e6fb7d4cc9459758ae/src/app/services/files.js#L122C9-L122C19) (delete from network + database) or standardize by marking files as "DELETED" so the background process takes care of these too? 

Since this endpoint is mainly used for backup files, direct deletion reduces database entries we might not use anymore (as they are for backups). But shouldn't we use the same background process for consistency? Drive-server is an old implementation so I just want to confirm if we should just mimic it for this case.

apsantiso (changes requested):

src/modules/file/file.repository.ts
```diff
@@ -752,6 +754,8 @@ export class SequelizeFileRepository implements FileRepository {
         removedAt: new Date(),
         status: FileStatus.DELETED,
         updatedAt: new Date(),
+        deleted: true,
+        deletedAt: new Date(),
```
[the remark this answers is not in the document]
evillalba94: @apsantiso I changed this because I assumed that a removed file would also be deleted, since removal is the final step in the process. But I'll reverse it.

src/modules/file/file.usecase.ts
sg-gs: Standardize as 'DELETED' better

apsantiso (approved):

PR #487, opened 2025-03-19
[PB-3705] Feat: update triggers for lookup table

evillalba94:
### Run script
- The trigger that allow to delete files and folders in the lookup table has been updated.
---
### Updates
- Update trigger to allow deleting files and folders in the lookUp table. The user is searched for by taking the information of the file and folder to be deleted. If the user is not searched for, a File/Folder cannot be deleted because `look_up.user_id` is a `UUID`, while `file.user_id` and `folder.user_id` are `INTEGER`.

### DIFF
--- a/migrations/20250311054626-update-triggers-for-lookup-table.js
+++ b/migrations/20250311054626-update-triggers-for-lookup-table.js
@@ -1,0 +1,72 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.sequelize.query(`
+      DROP TRIGGER IF EXISTS delete_file_from_look_up_table_after_file_deleted ON files;
+      DROP FUNCTION IF EXISTS delete_file_from_look_up_table();
+
+      DROP TRIGGER IF EXISTS delete_folder_from_look_up_table_after_folder_deleted ON folders;
+      DROP FUNCTION IF EXISTS delete_folder_from_look_up_table();
+    `);
+
+    await queryInterface.sequelize.query(`
+      CREATE OR REPLACE FUNCTION delete_file_from_look_up_table()
+      RETURNS TRIGGER AS $$
+      DECLARE
+        user_uuid VARCHAR;
+      BEGIN
+        SELECT uuid INTO user_uuid FROM users WHERE id = OLD.user_id;
+
+        -- Delete the lookup entry
+        DELETE FROM look_up
+        WHERE
+          item_type = 'file' 
+          AND item_id = OLD.uuid 
+          AND user_id = user_uuid;
+
+        RETURN OLD;
+      END;
+      $$ LANGUAGE 'plpgsql';
+
+      CREATE TRIGGER delete_file_from_look_up_table_after_file_deleted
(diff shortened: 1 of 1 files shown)


migrations/20250311054626-update-triggers-for-lookup-table.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.sequelize.query(`
+      DROP TRIGGER IF EXISTS delete_file_from_look_up_table_after_file_deleted ON files;
```
sg-gs, line 7: Remove the DROP, use CREATE or REPLACE, otherwise we risk to fail the second step and leave the production environment without any trigger at all.

migrations/20250311054626-update-triggers-for-lookup-table.js
```diff
@@ -0,0 +1,72 @@
...
+      CREATE TRIGGER delete_folder_from_look_up_table_after_folder_deleted
+      AFTER DELETE ON folders
+      FOR EACH ROW
+      EXECUTE FUNCTION delete_folder_from_look_up_table();
+    `);
+  },
+
+  async down(queryInterface, Sequelize) {
+    await queryInterface.sequelize.query(`
+      DROP TRIGGER IF EXISTS delete_file_from_look_up_table_after_file_deleted ON files;
```
sg-gs, line 65: This should be the CREATE or REPLACE of the previous trigger (look for it to copy-paste on the older migration), otherwise this is not undoing the migration but removing only the new trigger while leaving us with no trigger at all.

migrations/20250311054626-update-triggers-for-lookup-table.js
```diff
@@ -0,0 +1,72 @@
...
+        user_uuid VARCHAR;
+      BEGIN
+        SELECT uuid INTO user_uuid FROM users WHERE id = OLD.user_id;
+
+        -- Delete the lookup entry
+        DELETE FROM look_up
+        WHERE
+          item_type = 'file' 
+          AND item_id = OLD.uuid 
+          AND user_id = user_uuid;
```
sg-gs, line 27: The current function is: 

```
DELETE FROM look_up
WHERE
  item_type = 'file' 
  AND item_id = OLD.uuid;
```

I do not see how adding the `user_id` on the WHERE condition is useful at all (in fact, the SELECT before is adding more latency and binding us to an incremental ID which is precisely what we are trying to avoid due to the scalability issues that it may provoke when the possible IDs limit is reached). All the files have UUID, so I do not see how adding a less specific condition (as the UUID is unique and only assigned to a given file) adds or solves anything. I would understand this change if you were using the file.id + user.id because there are files with no UUID, but that is not the case. 

Reading the description of the PR it says: 
_If the user is not searched for, a File/Folder cannot be deleted because look_up.user_id is a UUID, while file.user_id and folder.user_id are INTEGER._

But we are not using that user_id INTEGER field at any moment, then I am guessing what is happening here

sg-gs (approved):

sg-gs:
Migration run. Merging

PR #489, opened 2025-03-20
[PB-4112]: feat/add workspace joined and left notification events

jzunigax2:
Add notification events for users leaving or joining a workspace

### DIFF
--- a/src/modules/workspaces/workspaces.usecase.ts
+++ b/src/modules/workspaces/workspaces.usecase.ts
@@ -75,7 +75,6 @@ import { SharingAccessTokenData } from '../sharing/guards/sharings-token.interfa
 import { FuzzySearchUseCases } from '../fuzzy-search/fuzzy-search.usecase';
 import { WorkspaceLog } from './domains/workspace-log.domain';
 import { TrashItem } from './interceptors/workspaces-logs.interceptor';
-import { StorageNotificationService } from '../../externals/notifications/storage.notifications.service';
 
 @Injectable()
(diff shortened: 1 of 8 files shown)


src/modules/workspaces/workspaces.usecase.ts
```diff
@@ -2819,6 +2848,12 @@ export class WorkspacesUsecases {
       user.uuid,
       workspaceId,
     );
+
+    this.storageNotificationsService.workspaceLeft({
+      payload: { workspaceId: workspace.id, workspaceName: workspace.name },
+      user,
+      clientId,
+    });
   }
```
apsantiso, line 2858: Do you have a way to implement the notification logic from the controllers instead of inserting it here? Just asking so we don't end up having to pass clientId through multiple usecases just to send notifications.

It doesn't matter if you have to fetch some rows again - this isn't a frequent or time-sensitive operation. That approach would be better than hiding this call here. 

It is not a must for all the functions, so if something seems to get too complex, you can just leave it inside the use case function

apsantiso (changes requested):

src/modules/workspaces/workspaces.usecase.ts
jzunigax2: I had that dilema, but wanted to avoid overcomplicating the controllers. Let me try refactoring the pr a bit

apsantiso (approved):

apsantiso (approved):

PR #532, opened 2025-04-29
feat(recover-account): recover account with legacy backup file 

apsantiso:
### Changes

- Added `legacy-recover-account` endpoint.
- WorkspaceUsers encrypted key of owned workspaces are updated with ecc encrypted mnemonic when account is recovered.
- User is removed from invited sharings and workspaces due to new keys generated. **We should warn users about this.**
- New generated keys are also saved

### DIFF
--- a/src/externals/crypto/crypto.service.ts
+++ b/src/externals/crypto/crypto.service.ts
@@ -22,7 +22,7 @@ export class CryptoService {
       this.configService.get('secrets.magicSalt'),
       this.configService.get('secrets.cryptoSecret2'),
     );
-    this.cryptoSecret = process.env.CRYPTO_SECRET;
+    this.cryptoSecret = this.configService.get('secrets.cryptoSecret');
   }
 
   encrypt(text: string, buffer?: Buffer) {
--- a/src/modules/user/user.usecase.spec.ts
+++ b/src/modules/user/user.usecase.spec.ts
@@ -82,6 +82,8 @@ import { WorkspacesUsecases } from '../workspaces/workspaces.usecase';
 import * as jwtLibrary from '../../lib/jwt';
 import { JsonWebTokenError } from 'jsonwebtoken';
 import { LegacyRecoverAccountDto } from './dto/legacy-recover-account.dto';
+import { CryptoModule } from '../../externals/crypto/crypto.module';
+import { AsymmetricEncryptionModule } from '../../externals/asymmetric-encryption/asymmetric-encryption.module';
 
 jest.mock('../../middlewares/passport', () => {
   const originalModule = jest.requireActual('../../middlewares/passport');
@@ -2644,12 +2646,8 @@ describe('User use cases', () => {
 
     beforeEach(async () => {
       const moduleRef: TestingModule = await Test.createTestingModule({
-        providers: [
-          UserUseCases,
-          CryptoService,
-          AsymmetricEncryptionService,
-          KyberProvider,
-        ],
+        providers: [UserUseCases],
+        imports: [CryptoModule, AsymmetricEncryptionModule],
       })
         .useMocker(() => createMock())
         .compile();
@@ -2670,9 +2668,8 @@ describe('User use cases', () => {
       const keys = await asymmetricEncryptionService.generateNewKeys();
       newCredentials = {
         mnemonic: mnemonicEncryptedWithPassword,
-        password:
-          '53616c7465645f5f48c4064882e75747c5db100136d111d08892ee5db2e1bb8c8d5dc91a0906fcaf0aa66dd13624e05736ef0d4e934e083a84ed7c1c82b7b64a8e51bad006ea18075bdcabcf0c1bd38cd47cca00c4be3cb9a794c4b87bf761be',
-        salt: '53616c7465645f5f0d114ce687f8728ea7db82dff0acea6bfd924275f74cc1c7725a70febf9caef83f2a49651122f8889bfb523513dec25a99910ebd339f15fd',
+        password: cryptoService.encryptText(decryptedPasswordHash),
+        salt: cryptoService.encryptText(decryptedSalt),
         asymmetricEncryptedMnemonic: {
           ecc: await asymmetricEncryptionService.hybridEncryptMessageWithPublicKey(
             {
@@ -2719,7 +2716,6 @@ describe('User use cases', () => {
       jest
         .spyOn(workspaceRepository, 'findWorkspaceUsersOfOwnedWorkspaces')
         .mockResolvedValue([]);
-
       await userUseCases.recoverAccountLegacy(user.uuid, newCredentials);
 
       expect(userRepository.updateByUuid).toHaveBeenCalledWith(user.uuid, {
--- a/src/modules/user/user.usecase.spec.ts
+++ b/src/modules/user/user.usecase.spec.ts
@@ -2680,7 +2680,6 @@ describe('User use cases', () => {
       jest
         .spyOn(workspaceRepository, 'findWorkspaceUsersOfOwnedWorkspaces')
         .mockResolvedValue([]);
-      jest.spyOn(userUseCases, 'resetUser').mockResolvedValue(undefined);
 
       await userUseCases.recoverAccountLegacy(userUuid, credentials);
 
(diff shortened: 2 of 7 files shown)


src/modules/user/user.usecase.ts
```diff
@@ -993,6 +993,102 @@ export class UserUseCases {
...
+    const { mnemonic, password, salt, asymmetricEncryptedMnemonic } =
+      credentials;
+
+    const user = await this.userRepository.findByUuid(userUuid);
+
+    if (!user) {
+      throw new NotFoundException('User not found');
+    }
+
+    await this.userRepository.updateByUuid(userUuid, {
```
sg-gs, line 1009: I would suggest to do this as the last step. Why? Because if the steps after this one fail, and even if the API returns an error, the door for signing in successfully when not everything it's okay will be open for the user.

As you already updated the password, the user could access to the account but its keys would be wrong. As this is a recover account operation, the user already cannot access to it, so it would be better to do this as the last step, as the user could sign in again only after completing every step successfully.

sg-gs (changes requested):

sg-gs:
Added @TamaraFinogina to this review as she has worked in this topic with us so having her feedback it's a stronger filter on these delicate details. @apsantiso

src/modules/user/user.usecase.spec.ts
```diff
@@ -2552,4 +2556,199 @@ describe('User use cases', () => {
...
+
+      jest.spyOn(userRepository, 'findByUuid').mockResolvedValue(user);
+      jest.spyOn(userRepository, 'updateByUuid').mockResolvedValue(undefined);
+      jest
+        .spyOn(cryptoService, 'decryptText')
+        .mockReturnValueOnce(decryptedPassword)
+        .mockReturnValueOnce(decryptedSalt);
+      jest
+        .spyOn(workspaceRepository, 'findWorkspaceUsersOfOwnedWorkspaces')
+        .mockResolvedValue([]);
```
TamaraFinogina, line 2665: Shall we test it with real crypto?

src/modules/user/user.usecase.spec.ts
```diff
@@ -2552,4 +2556,199 @@ describe('User use cases', () => {
...
+      ];
+
+      jest.spyOn(userRepository, 'findByUuid').mockResolvedValue(user);
+      jest.spyOn(userRepository, 'updateByUuid').mockResolvedValue(undefined);
+      jest.spyOn(cryptoService, 'decryptText').mockReturnValue('decrypted');
+      jest
+        .spyOn(workspaceRepository, 'findWorkspaceUsersOfOwnedWorkspaces')
+        .mockResolvedValue(workspaceAndUsers);
+      jest
+        .spyOn(workspaceRepository, 'updateWorkspaceUserEncryptedKeyByMemberId')
```
TamaraFinogina, line 2730: This one will be nice to test with real crypto too. The fact that all functions are called in the right order won't guarantee it decrypts because it misses the format/encoding/worng key/etc issues

src/modules/user/user.usecase.ts
```diff
@@ -993,6 +993,102 @@ export class UserUseCases {
...
+      hKey: this.cryptoService.decryptText(salt),
+    });
+
+    const keys = credentials.keys;
+    const keyVersions = Object.values(UserKeysEncryptVersions);
+
+    for (const version of keyVersions) {
+      const key = keys[version];
+      if (key) {
+        await this.keyServerUseCases.updateByUserAndEncryptVersion(
```
TamaraFinogina, line 1021: I think we have only 2 key versions:
```
export enum UserKeysEncryptVersions {
  Ecc = 'ecc',
  Kyber = 'kyber',
}

```
So, can it happen that we update only the ECC keys of a user who has both ECC and Kyber keys? Because if so, it will create problems. I'd say if we only have ECC keys, then Kyber has to be set to null. 

Also, can we even have a legacy account with Kyber keys?

TamaraFinogina: Wait, according to PB-4237, for legacy backups, the front always creates both keys. So, maybe this part should be changed a bit, because` if (key)` should be true for both. If not, then something unexpected happened.

src/externals/crypto/crypto.service.ts
```diff
@@ -22,7 +22,7 @@ export class CryptoService {
       this.configService.get('secrets.magicSalt'),
       this.configService.get('secrets.cryptoSecret2'),
     );
-    this.cryptoSecret = process.env.CRYPTO_SECRET;
+    this.cryptoSecret = this.configService.get('secrets.cryptoSecret');
```
apsantiso, line 25: This was breaking tests due to Nestjs not loading correctly this env value despite being present in `env.test`

apsantiso:
@TamaraFinogina @sg-gs I've added tests using real data (encrypting the password hash as we do in drive-web).

I'd like to emphasize that it would be better to handle this through integration or end-to-end tests rather than unit tests. However, unit tests are currently the only reliable way we have to ensure everything is working as expected so I took @TamaraFinogina suggestion 👍🏻

src/modules/user/user.usecase.ts
```diff
@@ -995,6 +995,105 @@ export class UserUseCases {
...
+    const ownedWorkspaceAndUsers =
+      await this.workspaceRepository.findWorkspaceUsersOfOwnedWorkspaces(
+        user.uuid,
+      );
+
+    if (ownedWorkspaceAndUsers.length > 0) {
+      //  TODO: this should be updated when we add support for workspace hybrid keys
+      await Promise.all(
+        ownedWorkspaceAndUsers.map((workspaceAndUser) =>
+          this.workspaceRepository.updateWorkspaceUserEncryptedKeyByMemberId(
```
sg-gs, line 1038: As the recovery is using the same mnemonic as it was used by the members when they joined the workspace, why is this needed? We use the public key of each member to encrypt the mnemonic of the owner. As the mnemonic does not change, why is this step performed?

TamaraFinogina: @sg-gs, it's not doing re-encryption. As I got it, it only sets the same `asymmetricEncryptedMnemonic.ecc` to all `memberId` of the workspace (changes table `workspace_users`, I assume), where `asymmetricEncryptedMnemonic.ecc` is an encryption of the workspace owner's mnemonic with the workspace owner's new keys. Because this is what the update function does:

```
async updateWorkspaceUserEncryptedKeyByMemberId(
    memberId: WorkspaceUserAttributes['memberId'],
    workspaceId: WorkspaceUserAttributes['workspaceId'],
    encryptedKey: WorkspaceUserAttributes['key'],
  ): Promise<void> {
    await this.modelWorkspaceUser.update(
      { key: encryptedKey },
      {
        where: { memberId, workspaceId },
      },
    );
  } 
```

Not sure why the workspace owner stores this data per `memberId`, it's better to ask @apsantiso. But it does it: 

-  During `workspaceUser` creation, `key` is set to `setupWorkspaceDto.encryptedMnemonic` :  https://github.com/internxt/drive-server-wip/blob/4d6700ab22b676176601e40a66ffe998b8d68031/src/modules/workspaces/workspaces.usecase.ts#L258

- And `encryptedMnemonic`is 'Owner mnemnonic encrypted with their public key in base64':
https://github.com/internxt/drive-server-wip/blob/4d6700ab22b676176601e40a66ffe998b8d68031/src/modules/workspaces/dto/setup-workspace.dto.ts#L30

As for the workspace invitees, you are right, they don't need to change anything as mnemonic is the same and their public keys haven't changed. But I think it's unrelated to this code.

sg-gs: The only case it will has sense, I think, is to update all the workspace_users where the user is the owner, as when the time to access to the workspace comes, the mechanism used to recover the workspace encryption key will check that (so the treatment of any member, including the owner, is the same when recovering its workspace keys). Let's see what @apsantiso has to say about it

apsantiso: @sg-gs Yes, exactly what you mentioned. It is to update the `workspace_users` where the user is the owner. We have a `workspace_user` for the owner too, who is managed like the rest of the members, so I perform this update so the owner is able to access the workspace without any problem with their new keys.

sg-gs (approved):

TamaraFinogina (approved):

PR #534, opened 2025-04-30
fix: add unique constraint to workspace_users table for workspace_id and member_id

jzunigax2:
There is currently a possible race condition on the setupWorkspace function, that may cause an owner to join their workspace twice. This is very similar to an error encountered a couple of months ago, in that case the race condition occured when accepting an invite. This Pr adds a unique constraint to workspaceUsers on the combination of memberId and workspaceId


jzunigax2:
Closing this as its a duplicate of #421

PR #564, opened 2025-05-28
[_]: chore(eslint): update rules to improve code quality

jzunigax2:
Adding some new eslint rules to help prevent common issues marked by sonar.

`prefer-readonly`: variables not declared as readonly was possibly the biggest number of issues present in the codebase.
`prefer-optional-chain`: instead of doing `payload && payload.uuid` we should opt for `payload?.uuid`
`consistent-type-imports`: would help prevent cyclic imports. most of the cyclic imports I found while reviewing sonar issues could be fixed by simply opting to import types


apsantiso:
This was merged from another PR

apsantiso (approved):

PR #584, opened 2025-06-17
[PB-4416]: feat(thumbnails)/add file_uuid column to thumbnails and populate exsting records

jzunigax2:
Add folderUuid field to thumbnails model as currentlythumbnails only relate to files using their numeric autoincremental ids. We are initially setting this field as nullable until we are able to populate existing records, this pr also introduces a script with this intention.

Based on the script in #542, we will attempt to populate thumbnails in batches, idempotently and with a retry mechanism

### DIFF
--- a/migrations/20250617204500-populate-fileuuids-in-thumbnails-batch.js
+++ b/migrations/20250617204500-populate-fileuuids-in-thumbnails-batch.js
@@ -1,0 +1,73 @@
+'use strict';
+
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
+const MAX_ATTEMPTS = 10;
+const BATCH_SIZE = 1000;
+const SLEEP_TIME_MS = 5000;
+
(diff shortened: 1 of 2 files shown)


migrations/20250617204500-populate-fileuuids-in-thumbnails-batch.js
```diff
@@ -0,0 +1,73 @@
...
+      'Starting migration: populating thumbnails with missing file_uuid',
+    );
+
+    do {
+      try {
+        const [results] = await queryInterface.sequelize.query(updateQuery);
+        updatedCount = results.length;
+        attempts = 0;
+
+        console.info(`Updated ${updatedCount} thumbnails in this batch`);
```
sg-gs, line 40: Use the accumulation (the total updated records) also please, thus I can follow the general progress

sg-gs (changes requested):

PR #593, opened 2025-06-30
[PB-4499]: Feature/reCaptcha for log in after some failed attempts

xabg2:
This PR introduces a new Guard that requires a CAPTCHA token after 5 failed login attempts. This helps prevent automated bots from brute-forcing user credentials.

### DIFF
--- a/src/modules/auth/conditional-captcha.guard.spec.ts
+++ b/src/modules/auth/conditional-captcha.guard.spec.ts
@@ -202,4 +202,5 @@ const buildUserWithErrorLoginCount = (count: number) => ({
   errorLoginCount: count,
   isGuestOnSharedWorkspace: () => false,
   toJSON: undefined,
+  hasBackupsEnabled: () => false,
 });


src/modules/auth/conditional-captcha.guard.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    const { errorLoginCount } = await this.userUseCase.findByEmail(email);
+
+    if (errorLoginCount >= this.threshold) {
+      if (!captchaToken) {
+        throw new UnauthorizedException(
+          'Captcha is required after multiple failed attempts',
+        );
+      }
```
apsantiso, line 38: Just as a note: we're using the same HTTP code for all three errors thrown here. If this means the app would need to rely on the message to know when to send the captcha, then we might want to consider including something extra in the returned error object.

apsantiso: If you really need to send the same http code, you could try this:

```
throw new UnauthorizedException({
  message: 'Captcha is required after multiple failed attempts',
  errorCode: 'CAPTCHA_REQUIRED',
  metadata: { threshold: this.threshold }
});
```

apsantiso (approved):

sg-gs (approved):

PR #601, opened 2025-07-01
[PB-4416]: refactor(sharing)/refactor file findFilesByOwnerAndSharedWithMe query to not rely on file id groupBy

jzunigax2:
findFilesByOwnerAndSharedWithMe used a groupBy that included the numeric fileId, since that is file model's primary key there was no easy way to simply exclude it from the query. So I ended up refactoring the business logic a bit performing to simple queries instead of the single more complex one that was being used.


src/modules/sharing/sharing.repository.ts
```diff
@@ -409,55 +409,51 @@ export class SequelizeSharingRepository implements SharingRepository {
...
       ],
-      order: orderBy,
-      limit,
-      offset,
     });
 
-    return sharedFiles.map((shared) => {
-      return this.toDomainFile(shared);
-    });
+    return allSharingsForItems.map((shared) => this.toDomainFile(shared));
```
apsantiso, line 456: My only concern here is... what if all files have 100+ records in sharings?

apsantiso (changes requested):

PR #604, opened 2025-07-03
[_] feat: added folders to devices

apsantiso:


sg-gs:
> key + platform + user should be unique

If I use the same platform, as the same user (same device case), how is the key guaranteeing the uniqueness @apsantiso ?

src/modules/backups/models/device.model.ts
```diff
@@ -34,6 +35,19 @@ export class DeviceModel extends Model implements DeviceAttributes {
   @Column
   userId: number;
 
+  @AllowNull
+  @ForeignKey(() => FolderModel)
+  @Column({
+    type: DataType.UUID,
+    field: 'folder_uuid',
+  })
+  folderUuid: string;
```
sg-gs, line 44: I assume this is required for the already existing ones, although is it better to do it this way or to use a fixed value for the already existing and not null in general?

src/modules/backups/models/device.model.ts
```diff
@@ -34,6 +35,19 @@ export class DeviceModel extends Model implements DeviceAttributes {
...
+
+  @AllowNull
+  @Index
+  @Column(DataType.STRING)
+  key: string;
```
sg-gs, line 49: Same here

PR #605, opened 2025-07-03
[PB-4504] feat: invalidate session after logout

apsantiso:
We need to start invalidating sessions when users log out. A JTI was added to the tokens in this PR https://github.com/internxt/drive-server-wip/pull/599. This PR intends to enable JWT blacklisting using Redis.

### Changes

- If a token contains a JTI and a request to /logout is made, the token used for the request will be invalidated until its expiration date

### DIFF
--- a/src/modules/cache-manager/cache-manager.service.ts
+++ b/src/modules/cache-manager/cache-manager.service.ts
@@ -8,7 +8,6 @@ export class CacheManagerService {
   private readonly LIMIT_KEY_PREFIX = 'limit:';
   private readonly JWT_KEY_PREFIX = 'jwt:';
   private readonly TTL_10_MINUTES = 10000 * 60;
-  private readonly TTL_7_DAYS = 604800;
 
   constructor(@Inject(CACHE_MANAGER) private readonly cacheManager: Cache) {}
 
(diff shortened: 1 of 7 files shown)


src/modules/cache-manager/cache-manager.service.ts
```diff
@@ -1,12 +1,14 @@
...
 import { CACHE_MANAGER } from '@nestjs/cache-manager';
 import { Cache } from 'cache-manager';
 
 @Injectable()
 export class CacheManagerService {
   private readonly USAGE_KEY_PREFIX = 'usage:';
   private readonly LIMIT_KEY_PREFIX = 'limit:';
+  private readonly JWT_KEY_PREFIX = 'jwt:';
   private readonly TTL_10_MINUTES = 10000 * 60;
+  private readonly TTL_7_DAYS = 604800;
```
sg-gs, line 11: I assume this 7 days are due to the max possible JWT duration allowed on the login. I recommend binding that value to the jwt sign one by using the same number in the [expiresIn](https://www.npmjs.com/package/jsonwebtoken) value when signing a token, as it also accepts a number

sg-gs (approved):

PR #607, opened 2025-07-07
[_]: feat(migrations)/add not-null constraint on file_uuid in thumbnails table

jzunigax2:
Add not null constraint to file_uuid field on thumbnails table


sg-gs (approved):
Running the migration.

PR #615, opened 2025-07-10
[_]: fix/empty-trash-giving-timeouts

larryrider:
- It optimizes the trash emptying process by batching deletions using Promise.all

### DIFF
--- a/src/modules/trash/trash.usecase.ts
+++ b/src/modules/trash/trash.usecase.ts
@@ -53,7 +53,10 @@ export class TrashUseCases {
       );
     }
 
-    await Promise.all(deleteItemsChunks);
+    const promiseChunkSize = 10;
+    for (let i = 0; i < deleteItemsChunks.length; i += promiseChunkSize) {
+      await Promise.all(deleteItemsChunks.slice(i, i + promiseChunkSize));
+    }
   }
 
   /**


src/modules/trash/trash.usecase.ts
```diff
@@ -23,25 +23,39 @@ export class TrashUseCases {
     const foldersCount = await this.folderUseCases.getTrashFoldersCount(
       trashOwner.id,
     );
-    const emptyTrashChunkSize = 100;
+
+    const emptyTrashChunkSize = 1000;
```
sg-gs, line 27: Be careful @larryrider, the low numbers are intentionally used to prevent scalability issues. Which are the benchmarks in this case?

larryrider: Good point, I haven't done proper benchmarking yet.
This change came up because users with a lot of files/folders in trash are experiencing timeouts when calling the endpoint. The previous chunk size (100) is processing things too slowly and not finishing in time. I bumped it up to 1000 as a quick fix to avoid the request failing.

After chatting with Andrés, we added too a Promise.all in batches to speed things up, so maybe we can now actually keep the original chunk size and just see if this batch processing already solves the timeout issue.

The tricky part I think is that testing this locally isn't super realistic as the benchmark in dev is nothing like what happens in prod with thousands of items. Not really sure what the best way to benchmark this would be tbh. Open to ideas

PR #640, opened 2025-07-31
[PB-4737]: fix/match user UUID in backupfile with UUID in token

jzunigax2:
When recovering an account match the user's uuid in the recovery token with the uuid optionally provided in the body. This is matched if and only if the uuid was provided, this is assuming the new backup recovery format will go into production as is, (without specifying user uuid) and the pr that introduces this field to the file will be merged later on

### DIFF
--- a/src/modules/user/dto/recover-account.dto.ts
+++ b/src/modules/user/dto/recover-account.dto.ts
@@ -25,8 +25,8 @@ export class RecoverAccountDto {
     example: '123e4567-e89b-12d3-a456-426614174000',
     description: 'User uuid',
   })
-  @IsNotEmpty()
-  uuid: string;
+  @IsOptional()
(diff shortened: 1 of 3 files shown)


src/modules/user/dto/recover-account.dto.ts
```diff
@@ -21,6 +21,13 @@ export class RecoverAccountDto {
   @IsNotEmpty()
   password: string;
 
+  @ApiProperty({
+    example: '123e4567-e89b-12d3-a456-426614174000',
+    description: 'User uuid',
+  })
+  @IsOptional()
+  uuid?: string;
+
```
apsantiso, line 30: If this is going to be used for the new endpoint and backup key file and the frontend is not merged yet, force it.

PR #644, opened 2025-08-05
[_] feat: add retroactive cleanup script

apsantiso:
This is a script that handles any undeleted files or folders that the background process may have missed. Unlike the [cron job](https://github.com/internxt/drive-server-wip/pull/638), which only handles folders that are missed starting from the day it is merged, this script also takes care of older, missed folders.

### How to run
```
yarn run cleanup:retroactive

// In case the process stopped and you need to start from a specific user id
yarn run cleanup:retroactive --userId=numericId
````

### Flow

1. Fetch users in batches, ordered by ID.
2. For each batch of users, find the last deleted folder for each user. This approach saves time by avoiding a scan of all folders, which is particularly efficient for users who do not frequently delete folders.
3. If a deleted folder is found, retrieve the `update_at` field as that means the most recent deletion date.
4. This most recent deletion date serves as the upper limit for our scan. We do not need to scan for folders that are going to be deleted by this script. Instead, we use this date to find folders that were missed by drive-background-tasks. **If we mark a folder with children as deleted, it will be handled by either the cron job or drive-background-tasks.**
5. Scan for and retrieve the UUIDs of deleted parent folders with existent children.
6. Update all child folders within these parent folders that have not yet been marked as deleted. The query used is `UPDATE SET ... WHERE parent_uuid IN (....) AND removed = false`, which ensures that already deleted folders are not affected.
7. After processing all deleted parent folders up to the date retrieved in step 4, we proceed to clean up any missed files.
8. Scan the database for files marked as EXISTS within the user's deleted folders.
9. Update any file inside the deleted folders that has not been updated yet.
10. Repeat this process until all users have been processed.

### What if there's any error?
You can take a look at the logs to check the id of the last processed user and start the script again `yarn run cleanup:retroactive --userId=numericId`

image

### DIFF
--- a/src/modules/folder/folder.repository.ts
+++ b/src/modules/folder/folder.repository.ts
@@ -988,7 +988,7 @@ export class SequelizeFolderRepository implements FolderRepository {
     const lastDeletedFolder = await this.folderModel.findOne({
       where: {
         userId,
-        removed: false,
+        removed: true,
       },
       limit: 1,
       order: [['updated_at', 'DESC']],

### CODE REVIEW

File: src/modules/folder/folder.repository.ts [hunk 1]
```diff
@@ -978,6 +984,77 @@ export class SequelizeFolderRepository implements FolderRepository {
     return { updatedCount };
   }
 
+  async getLastDeletedFolder(userId: number): Promise<Folder | null> {
+    const lastDeletedFolder = await this.folderModel.findOne({
+      where: {
+        userId,
+        removed: false,
```
sg-gs, line 991: If it is the last deleted, shouldn't this be true?

sg-gs (changes requested):

src/modules/file/file.repository.ts
```diff
@@ -730,6 +730,28 @@ export class SequelizeFileRepository implements FileRepository {
...
+      {
+        removed: true,
+        removedAt: deletedDate,
+        status: FileStatus.DELETED,
+        updatedAt: deletedDate,
+      },
+      {
+        where: {
+          folderUuid: { [Op.in]: parentUuids },
+          removed: false,
```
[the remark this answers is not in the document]
apsantiso: I was too focused on the folder's behavior that I missed this. Fixed, thanks!

apsantiso:
@sg-gs This PR was not synced with master, so it seems like I added a lot of commits after your last review, but it was just a rebase to fix the conflicts + the commit that fixes the file deletion issue you mentioned earlier.

src/modules/file/file.repository.ts
```diff
@@ -744,7 +744,7 @@ export class SequelizeFileRepository implements FileRepository {
       {
         where: {
           folderUuid: { [Op.in]: parentUuids },
-          removed: false,
+          status: FileStatus.DELETED,
```
sg-gs, line 747: Shouldn't this be ne? (!=)

sg-gs (changes requested):

sg-gs (approved):

PR #668, opened 2025-08-20
[PB-2394] feat: calculate yesterday usage and first usage on first /usage call

apsantiso:
⚠️ **Important:** This PR implements the calculation logic but does NOT change the response returned to users. The /usage endpoint continues to return the existing implementation.

Implements incremental usage calculation logic to track user storage usage over time. This PR adds the calculation and delta storage functionality without changing the current user-facing response, allowing us to monitor the new system's behavior before switching over.

## How a request to`/usage` works

1. **Find last usage record**: Query the latest usage record by `period` field (`last_usage`)
2. **Handle first-time users**: If no usage record exists, calculate delta from user creation date to yesterday and create initial usage record, then skip to step 4
3. **Handle existing users**: If `last_usage` exists, check if `last_usage.period === currentDate - 1 day`:
  - **Yes**: Proceed to step 4
  - **No**: Calculate delta from `last_usage.period` to `currentDate - 1 day`, save this delta in usage table with yesterday's date
4. **Calculate historical usage**: Sum all deltas from usage table **(PENDING)**
5. **Calculate today's usage**: Sum current day delta directly from files table **(PENDING)**
6. **Return total**: Sum both results and return to user **(PENDING)**

## How replacing a file works

1. **File update**: The file is replaced successfully
2. **Did file size changed?:** We calculate delta by comparing the old file size and the new file size
3. **Save delta if needed:** If there's a change, we store delta with daily type

### **Why are files created and updated on the same day ignored?**

These files will be included in the next delta calculation (the following day). If we also save a delta for them on the current day, it would result in double-counting and incorrect calculations.

### DIFF
--- a/src/modules/file/file.repository.ts
+++ b/src/modules/file/file.repository.ts
@@ -125,7 +125,7 @@ export interface FileRepository {
     order?: [keyof FileModel, 'ASC' | 'DESC'][],
   ): Promise<File[]>;
   deleteUserTrashedFilesBatch(userId: number, limit: number): Promise<number>;
-  sumFileSizesChangesBetweenDates(
+  sumFileSizeDeltaBetweenDates(
     userId: FileAttributes['userId'],
     sinceDate: Date,
     untilDate?: Date,
@@ -899,7 +899,7 @@ export class SequelizeFileRepository implements FileRepository {
     return Number(result[0]['total']) as unknown as number;
   }
 
-  async sumFileSizesChangesBetweenDates(
+  async sumFileSizeDeltaBetweenDates(
     userId: FileAttributes['userId'],
     sinceDate: Date,
     untilDate?: Date,
--- a/src/modules/file/file.repository.ts
+++ b/src/modules/file/file.repository.ts
@@ -940,7 +940,6 @@ export class SequelizeFileRepository implements FileRepository {
         ],
       },
       raw: true,
-      logging: true,
     });
 
     return Number(result[0]['total']) as unknown as number;
--- a/src/modules/file/file.repository.ts
+++ b/src/modules/file/file.repository.ts
@@ -125,6 +125,11 @@ export interface FileRepository {
     order?: [keyof FileModel, 'ASC' | 'DESC'][],
   ): Promise<File[]>;
   deleteUserTrashedFilesBatch(userId: number, limit: number): Promise<number>;
+  sumFileSizesChangesBetweenDates(
+    userId: FileAttributes['userId'],
+    sinceDate: Date,
+    untilDate?: Date,
+  ): Promise<number>;
 }
 
 @Injectable()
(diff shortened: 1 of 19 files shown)


src/modules/file/file.repository.ts
```diff
@@ -125,6 +125,11 @@ export interface FileRepository {
     order?: [keyof FileModel, 'ASC' | 'DESC'][],
   ): Promise<File[]>;
   deleteUserTrashedFilesBatch(userId: number, limit: number): Promise<number>;
+  sumFileSizesChangesBetweenDates(
```
sg-gs, line 128: Let's use the 'delta' wording as it is more clear to understand

src/modules/file/file.repository.ts
```diff
@@ -893,4 +898,50 @@ export class SequelizeFileRepository implements FileRepository {
...
+      ...(untilDate ? { [Op.lte]: untilDate } : null),
+    };
+
+    const result = await this.fileModel.findAll({
+      attributes: [
+        [
+          Sequelize.literal(`
+            SUM(
+              CASE 
+                WHEN status = 'DELETED' AND date_trunc('day', created_at) = date_trunc('day', updated_at) THEN 0
```
sg-gs, line 918: If we choose from 3 days ago to now, the record will appear twice? (one for its creation and one for the removal?) Otherwise a file created yesterday and removed today will count as a negative delta, when it should be 0. Am I missing something?

src/modules/file/file.usecase.spec.ts
```diff
@@ -1443,4 +1448,108 @@ describe('FileUseCases', () => {
...
+      jest.setSystemTime(today);
+
+      jest
+        .spyOn(usageService, 'getUserMostRecentUsage')
+        .mockResolvedValue(null);
+      jest
+        .spyOn(usageService, 'createFirstUsageCalculation')
+        .mockResolvedValue(mockFirstUsage);
+
+      await service.getUserUsedStorageIncrementally(mockUser);
```
sg-gs, line 1479: What about the return, should you assert the returned value?

src/modules/file/file.usecase.spec.ts
```diff
@@ -1443,4 +1448,108 @@ describe('FileUseCases', () => {
...
+      });
+
+      // Set today to the next period start date according to mockUsage
+      jest.setSystemTime(today);
+
+      jest
+        .spyOn(usageService, 'getUserMostRecentUsage')
+        .mockResolvedValue(mockUsage);
+
+      await service.getUserUsedStorageIncrementally(mockUser);
```
sg-gs, line 1504: What about the return value?

src/modules/file/file.usecase.spec.ts
```diff
@@ -1443,4 +1448,108 @@ describe('FileUseCases', () => {
...
+        .spyOn(usageService, 'getUserMostRecentUsage')
+        .mockResolvedValue(mockUsage);
+      jest
+        .spyOn(fileRepository, 'sumFileSizesChangesBetweenDates')
+        .mockResolvedValue(mockGapDelta);
+      jest
+        .spyOn(usageService, 'createMonthlyUsage')
+        .mockResolvedValue(undefined);
+
+      await service.getUserUsedStorageIncrementally(mockUser);
```
sg-gs, line 1536: What about the return value?

sg-gs (changes requested):

PR #671, opened 2025-08-21
Feat/incremental usage 2

apsantiso:


src/modules/usage/usage.repository.ts
```diff
@@ -65,6 +65,105 @@ export class SequelizeUsageRepository {
...
+        WITH yearly_sums AS (
+          SELECT
+              date_trunc('year', period) AS year,
+              SUM(delta) AS total_delta
+          FROM
+              public.usages
+          WHERE
+              type = 'yearly'
+              AND user_id = :userUuid
+          GROUP BY
```
sg-gs, line 79: Why are you grouping by year if it is a yearly record? The record itself is an aggregation. Am I missing something?

src/modules/usage/usage.repository.ts
```diff
@@ -65,6 +65,105 @@ export class SequelizeUsageRepository {
...
+                  date_trunc('year', period) AS year,
+                  date_trunc('month', period) AS month,
+                  SUM(delta) AS total_delta
+              FROM
+                  public.usages
+              WHERE
+                  type = 'monthly'
+                  AND user_id = :userUuid
+              GROUP BY
+                  date_trunc('year', period), date_trunc('month', period)
```
sg-gs, line 93: Same question here

sg-gs, line 93: Same question here

src/modules/usage/usage.repository.ts
```diff
@@ -65,6 +65,105 @@ export class SequelizeUsageRepository {
...
+                  date_trunc('year', period) AS year,
+                  date_trunc('month', period) AS month,
+                  SUM(delta) AS total_delta
+              FROM
+                  public.usages
+              WHERE
+                  type = 'daily'
+                  AND user_id = :userUuid
+              GROUP BY
+                  date_trunc('year', period), date_trunc('month', period)
```
sg-gs, line 106: And here, regarding to the year

src/modules/usage/usage.repository.ts
```diff
@@ -84,6 +84,55 @@ export class SequelizeUsageRepository {
...
+        WHERE type = 'yearly' AND user_id = :userUuid
+    ),
+    aggregated_data AS (
+        -- Aggregate yearly data where it exists
+        SELECT
+            date_trunc('year', period) AS year,
+            SUM(delta) AS total_delta
+        FROM public.usages
+        WHERE type = 'yearly' AND user_id = :userUuid
+        GROUP BY date_trunc('year', period)
```
apsantiso, line 101: `period` is not aggregated, so we need to either group by it or use an aggregate function on it, as we want to get the year field, then we just group by the field.

PR #679, opened 2025-08-25
[_] fix(usage): add unique constraint to usage table

apsantiso:
This adds an unique constraint using INDEX as postgresql does not allow partial constraints by default. When two calls to `/usage` are made at the same time, a race condition is created, therefore, two records are created in the table with the same value.

⚠️ We need to clean the usages table before running the index creation !!


sg-gs:
Deleted everything from the `usages` table @apsantiso 
image

sg-gs (approved):
Running migration

PR #684, opened 2025-08-29
Feat/notifications kickoff

apsantiso:


src/modules/notifications/notifications.module.ts
```diff
@@ -0,0 +1,30 @@
...
+      NotificationModel,
+      UserNotificationStatusModel,
+    ]),
+  ],
+  controllers: [NotificationsController],
+  providers: [
+    {
+      provide: NotificationRepository,
+      useClass: SequelizeNotificationRepository,
+    },
```
apsantiso, line 25: We use an abstract class (acting as interface) to provide the implementation. Interfaces are not compiled so we use an abstract class instead.

This is the approach we should've used in other repositories so we do not use the implementation as type in the services (eg: file.repository.ts).

src/modules/notifications/models/notification.model.ts
```diff
@@ -0,0 +1,50 @@
...
+})
+export class NotificationModel extends Model implements NotificationAttributes {
+  @PrimaryKey
+  @Default(DataType.UUIDV4)
+  @Column(DataType.UUID)
+  id: string;
+
+  @AllowNull(false)
+  @Column(DataType.TEXT)
+  link: string | null;
```
sg-gs, line 25: If allow null is false, this type is wrong

sg-gs (approved):

PR #690, opened 2025-09-02
feat(notifications): added create endpoint

apsantiso:
Changes
- Added notification creation endpoint
- Some endpoints should be protected so they're not accessible by users, added a guard with a hardcoded false return to prevent any request to the endpoint


src/modules/notifications/notifications.controller.ts
```diff
@@ -1,11 +1,39 @@
...
+    description: 'Notification created successfully',
+    type: NotificationResponseDto,
+  })
+  async createNotification(
+    @Body() createNotificationDto: CreateNotificationDto,
+  ): Promise<NotificationResponseDto> {
+    const notification = await this.notificationsUseCases.createNotification(
+      createNotificationDto,
+    );
+    return new NotificationResponseDto(notification);
```
apsantiso, line 37: This enables [serialization](https://docs.nestjs.com/techniques/serialization), so we do not give the consumers fields such as targetType, targetValue, they just know the fields we want them to know.

sg-gs (approved):

PR #701, opened 2025-09-10
[PB-4832]  feat(cli): added login for cli

apsantiso:


src/modules/feature-limit/feature-limit.service.ts
```diff
@@ -0,0 +1,38 @@
...
+      this.logger.warn(
+        {
+          tierId,
+          platform,
+          category: 'MISSING-LIMIT',
+        },
+        'Missing platform access limit for tier, bypassing access check',
+      );
+      return true;
+    }
```
apsantiso, line 34: I added a bypass here. We could simply prevent access, but what if we create a new plan and forget to map it? Users would complain about not being able to access features despite being paid users.

Anyway, we need to do a small rework of the tiers system so that Drive is the one that creates its own tiers, and payments just assigns them to users. So the probability of this issue happening is very low tbh

src/modules/auth/jwt.strategy.ts
```diff
@@ -69,11 +72,17 @@ export class JwtStrategy extends PassportStrategy(Strategy, strategyId) {
...
-        return this.userUseCases.getUserByUsername(user.bridgeUser);
+        //  Legacy shared workspaces. It is not the current workspaces implementation.
+        const sharedWorkspaceHost = await this.userUseCases.getUserByUsername(
+          user.bridgeUser,
+        );
+        return [sharedWorkspaceHost, authInfo];
       }
 
-      return user;
+      return [user, authInfo];
```
apsantiso, line 85: Got this from NestJs Passport abstraction [doc](https://docs.nestjs.com/recipes/passport#implementing-passport-jwt). We can access the authInfo using `req.authInfo`

> Additionally, you can return an array, where the first value is used to create a user object and the second value is used to create an authInfo object.

src/modules/feature-limit/feature-limit.service.ts
sg-gs: It's always good to take these details into consideration, so it's fine.

sg-gs (approved):

PR #724, opened 2025-09-26
[PB-5032] Add events module with initial payment processing support

douglas-xt:
This adds a events module to process specific controlled events from the frontend, for example payment processing.

- The endpoint **/events/payments/incomplete-checkout** is exposed for handling incomplete checkout events
- Receives the needed param *completeCheckoutUrl* for the template to send the email.
- Add a new method in mailer service async **sendIncompleteCheckoutEmail** and respective templateId env var.

### DIFF
--- a/src/modules/events/events.module.ts
+++ b/src/modules/events/events.module.ts
@@ -1,0 +1,13 @@
+import { Module } from '@nestjs/common';
+import { ConfigModule } from '@nestjs/config';
+import { MailerService } from '../../externals/mailer/mailer.service';
+import { EventsController } from './events.controller';
+import { EventsUseCases } from './events.usecase';
+
+@Module({
+  imports: [ConfigModule],
+  controllers: [EventsController],
+  providers: [MailerService, EventsUseCases],
+  exports: [EventsUseCases],
+})
+export class EventsModule {}
--- a/src/modules/events/events.usecase.ts
+++ b/src/modules/events/events.usecase.ts
@@ -1,0 +1,32 @@
+import { Injectable, Logger, Inject } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import { MailerService } from '../../externals/mailer/mailer.service';
+import { IncompleteCheckoutDto } from './dto/incomplete-checkout.dto';
+import { User } from '../user/user.domain';
+
+@Injectable()
+export class EventsUseCases {
+  constructor(
+    private readonly mailerService: MailerService,
+    @Inject(ConfigService)
+    private readonly configService: ConfigService,
+  ) {}
+
+  async handleIncompleteCheckoutEvent(
+    user: User,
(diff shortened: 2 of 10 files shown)


src/modules/events/events.module.ts
```diff
@@ -0,0 +1,13 @@
...
+import { EventsController } from './events.controller';
+import { EventsUseCases } from './events.usecase';
+
+@Module({
+  imports: [ConfigModule],
+  controllers: [EventsController],
+  providers: [MailerService, EventsUseCases],
+  exports: [EventsUseCases],
+})
+export class EventsModule {}
```
apsantiso, line 13: @douglas-xt Let's reduce the code we'll need to maintain, we can just put all this logic under the user's module/controller/usecase.

We'll probably never use an events module (at least not for this) as we use a DDD approach and everything is encapsulated in each domain. We do not have a payments module (not drive's responsibility) so you can use the users module for this.

IMO, any payment thing/mail should be sent by payments, but let's be pragmatic in this case.

src/modules/events/events.usecase.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  async handleIncompleteCheckoutEvent(
+    user: User,
+    dto: IncompleteCheckoutDto,
+  ): Promise<{ success: boolean }> {
+    try {
+      await this.mailerService.sendIncompleteCheckoutEmail(
+        user.email,
+        dto.completeCheckoutUrl,
+      );
```
apsantiso, line 23: Add a control for how many times per day we can send this email. If we don't implement any control, malicious users could spam this endpoint and exhaust our SendGrid daily quota. 

Check the security module -> mail limits module. Set this to something like a maximum of 5 emails per day, you'll find some examples across the entire codebase that you can follow.

apsantiso (changes requested):

apsantiso:
@sg-gs can you help us to run this migration please? Please @douglas-xt talk with Sergio about the .env that he needs to set so this endpoint works.

src/modules/user/user.usecase.ts
```diff
@@ -1951,4 +1952,50 @@ export class UserUseCases {
...
+      await this.mailLimitRepository.updateByUserIdAndMailType(
+        user.id,
+        MailTypes.IncompleteCheckout,
+        mailLimit,
+      );
+
+      return { success: true };
+    } catch (error) {
+      new Logger('[USER/INCOMPLETE_CHECKOUT]').error(
+        `Failed to send incomplete checkout email to ${user.email}: ${error.message}`,
```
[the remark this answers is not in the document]
apsantiso: You can let the global exception filter handle this for you. Any unexpected error would be handled and logged by it @douglas-xt

sg-gs (approved):

apsantiso (approved):

apsantiso (approved):

PR #728, opened 2025-09-30
[PB-4689] feat(audit-logs): kickoff module

apsantiso:
This is a branch off https://github.com/internxt/drive-server-wip/pull/633. 

### Changes
- Added audit logs module
- Added audit logs table creation

PR #729, opened 2025-09-30
[PB-4689] feat/audit logs decorator and service

apsantiso:
Dependency: https://github.com/internxt/drive-server-wip/pull/728

This PR adds a decorator that would let us to audit any endpoint by just setting it up correctly.

We'll decorate the controllers in another PR.

### DIFF
--- a/src/common/audit-logs/interceptors/audit-log.interceptor.spec.ts
+++ b/src/common/audit-logs/interceptors/audit-log.interceptor.spec.ts
@@ -10,7 +10,7 @@ import {
   AuditEntityType,
   AuditPerformerType,
 } from '../audit-logs.attributes';
-import { newUser } from '../../../../test/fixtures';
+import { newUser, newWorkspace } from '../../../../test/fixtures';
 
 describe('AuditLogInterceptor', () => {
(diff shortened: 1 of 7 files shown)


src/common/audit-logs/interceptors/audit-log.interceptor.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+
+      reflector.get.mockReturnValue(metadata);
+      auditLogService.log.mockResolvedValue(undefined);
+
+      await lastValueFrom(interceptor.intercept(context, next));
+
+      expect(auditLogService.log).toHaveBeenCalledWith({
+        action: AuditAction.WorkspaceCreated,
+        entityType: AuditEntityType.Workspace,
+        entityId: 'workspace-456',
```
sg-gs, line 104: Using `response.workspace.id` / `request.body.creatorId` / `request.body.name` to assert, instead of hardcoded values, improves the maintainability of this test

sg-gs (changes requested):

sg-gs (approved):

PR #731, opened 2025-10-02
[_] chore:add cli access limit

apsantiso:
### Changes
- Created migration to add cli-access to the limits table
- Created migration to allow cli access to all paid tiers


sg-gs (approved):
Running migration

PR #732, opened 2025-10-03
[PB-5092] feat(gateway): update workspace endpoint

apsantiso:


src/modules/gateway/gateway.usecase.spec.ts
```diff
@@ -592,5 +592,221 @@ describe('GatewayUseCases', () => {
...
+          });
+
+          expect(featureLimitService.getTier).toHaveBeenCalledWith(tier.id);
+          expect(userRepository.updateBy).toHaveBeenCalledWith(
+            { uuid: workspaceUser.uuid },
+            { tierId: tier.id },
+          );
+        });
+
+        it('When tier exists and is same as workspace user tier, then it should not update tier', async () => {
```
sg-gs, line 724: This test is missing the check to ensure the `updateBy` function has not been called

sg-gs (changes requested):

xabg2 (approved):

xabg2 (approved):

PR #735, opened 2025-10-07
[_] feat: expose user public info gateway

apsantiso:
### Changes
- Allow get users credentials via gateway


sg-gs:
Okay, merge it @apsantiso

PR #898, opened 2026-01-22
[PB-5673] set name = plainName

TamaraFinogina:
This PR sets the name equal to plainName instead of its encrypted version.


sg-gs (changes requested):
Left some comments @TamaraFinogina

sg-gs (commented):
LGTM, although do all the clients comply / expect / can manage this change ? Ask to both teams @TamaraFinogina

PR #972, opened 2026-03-04
[_] test_plainName

TamaraFinogina:
This PR attempts to return plainName instead of name to see if we can drop legacy name encryption/decryption.


jzunigax2 (approved):

sg-gs (commented):
Shouldn't we make `plainName` NOT NULL first @apsantiso ?

apsantiso:
> Shouldn't we make `plainName` NOT NULL first @apsantiso ?

We can't, there are some files without name. For example, macOS client uploads . DS_Store constanly

apsantiso:
@TamaraFinogina marked this as "blocked" so it does not get closed automatically.

PR #1019, opened 2026-03-31
[PB-1960]: feat/implement mail account provisioning and recovery email functionality

jzunigax2:
- Introduced Mail module with MailService and MailController for handling mail account creation.
- Implemented MailUseCases to manage mail account provisioning and validation.
- Updated user model to support recovery email and added necessary DTOs for mail account creation.
- Enhanced audit logging to track mail setup actions.
- Attempting a log in with an internxt owned email domain would promt an address search on mail-server


apsantiso (changes requested):

src/modules/mail/mail.usecase.ts
```diff
@@ -0,0 +1,104 @@
...
+    @Inject(FeatureLimitService)
+    private readonly featureLimitService: FeatureLimitService,
+  ) {}
+
+  async createMailAccount(
+    user: User,
+    dto: CreateMailAccountDto,
+  ): Promise<{ token: string; newToken: string; address: string }> {
+    await this.assertMailAccessEnabled(user);
+    this.verifyPassword(user, dto.password);
```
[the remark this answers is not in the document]
apsantiso: We can also share the token at some extend like we did with VPN @jzunigax2. Just throwing ideas tho. Not sure if that aligns with what @sg-gs thought

src/modules/mail/mail.usecase.ts
```diff
@@ -0,0 +1,104 @@
...
+      });
+      throw error;
+    }
+
+    const updatedUser = await this.userRepository.findByUuid(user.uuid);
+
+    const { token, newToken } =
+      await this.userUseCases.getAuthTokens(updatedUser);
+
+    return { token, newToken, address: fullAddress };
```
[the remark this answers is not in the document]
sg-gs: > Anyway, we need to think about the downgrade process. We would need to prevent new users from signing up with the recovery_email too.

Isn't this already handled by the signup itself? Or am I missing something? 

> Ok, what about preserving the email field as it is and store the internxt email in another field or even table (I would go for the second option to support multiple emails in the future)?
> 
> If the user wish to change their primary email they can always change the email manually.

At our level (before Stalwart) I assume we should have the _local-part_ (this naming comes from https://datatracker.ietf.org/doc/html/rfc5322) multiplexing per user on another table. 

> The only downside I see is related to any sharing/workspace operation that requires searching by mail. We would need to change the mail search in those usecases

Why? I mean, we will search by the local-part and from there, get the user (in case you need the search to find the user) or are you referring to something else?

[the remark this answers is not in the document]
sg-gs: > @sg-gs I mean we should change how we search for the user in the "invite user" usecase, which uses the email address to find the user. Not a big deal as we use mainly uuids tho. If I can login with [john@internxt.com](mailto:john@internxt.com) too, I would expect other users to find me as [john@internxt.com](mailto:john@internxt.com) too.

Let's assume this happens, in both ways: 
1. An Internxt user that has multiple mail extensions is invited to an item. Receives the email, joins with a token, Drive's session is the same, it works. 
     a. The internal invitation notifications will still appear as it is invited using its uuid. (which mail would have also)
2. An Internxt User that does not have Mail --> not affected. 

So I do not see any issue here 

> That's why I jumped in, as changing the main email would lead to weird things
This could be visual. I won't change it internally. Just do the following: 
1. User without mail -> works as always. 
2. User using a internxt mail domain -> ask Mail for the uuid, get the password on Drive with that uuid -> match password. 
And Drive would still have the same thing: email, password, uuid
And Mail would store: uuid, mail.localParts.userId -> mail.users.uuid (FK), mail.localParts.value (here, there would be as many rows per user as different domains it can have)

jzunigax2:
@sg-gs @apsantiso addressed concerns, removed recovery mail stuff and the email replacement logic. I'll improve test coverage, but in the meantime does it look good overall?

migrations/20260330160219-add-mail-access-limit.js
sg-gs, line 1: I think this should not be part of Drive, rather than from payments.

src/modules/auth/auth.controller.ts
```diff
@@ -60,8 +62,19 @@ export class AuthController {
...
+    private readonly mailService: MailService,
   ) {}
 
+  private async resolveLoginEmail(email: string): Promise<string> {
+    if (!isManagedMailDomain(email)) return email;
+
+    const userId = await this.mailService.findUserIdByAddress(email);
+    if (!userId) return email;
+
+    const user = await this.userUseCases.findByUuid(userId).catch(() => null);
```
sg-gs, line 74: Don't hide errors, log or let them propagate

src/modules/auth/auth.controller.ts
```diff
@@ -60,8 +62,19 @@ export class AuthController {
     private readonly twoFactorAuthService: TwoFactorAuthService,
     private readonly authUseCases: AuthUsecases,
     private readonly featureLimitService: FeatureLimitService,
+    private readonly mailService: MailService,
   ) {}
 
+  private async resolveLoginEmail(email: string): Promise<string> {
```
sg-gs, line 68: This is part of the login logic, I will add it to the auth.service, not in the controller, as this applies to the login itself, with independence of the web login or the CLI one

src/modules/mail/mail.controller.ts
```diff
@@ -0,0 +1,30 @@
+import { Body, Controller, HttpCode, HttpStatus, Post } from '@nestjs/common';
+import { ApiBearerAuth, ApiOkResponse, ApiOperation } from '@nestjs/swagger';
+import { User as UserDecorator } from '../auth/decorators/user.decorator';
+import { User } from '../user/user.domain';
+import { MailUseCases } from './mail.usecase';
+import { CreateMailAccountDto } from './dto/create-mail-account.dto';
+import { AuditLog } from '../../common/audit-logs/decorators/audit-log.decorator';
+import { AuditAction } from '../../common/audit-logs/audit-logs.attributes';
+
+@Controller('mail')
```
sg-gs, line 10: Why is this here? Shouldn't this be on mail-server?

jzunigax2:
closing in favor of https://github.com/internxt/mail-server/pull/25

PR #1029, opened 2026-04-08
chore/move to npm

apsantiso:
Yarn v1 has been in maintenance mode for a while now. npm has addressed most of its historical issues, so it's safe to move back to the default option.
## Changes

- Removed yarn.lock
- Migrated to npm
- Changed resolutions to overrides for npm compatibility
- Removed deprecated husky boilerplate
- Aligned node types with the version used in the Dockerfiles
- Added --ignore-scripts where missing
- Fixed jest version as using 30.3.0 leads to type errors. https://github.com/jestjs/jest/issues/16028. Let's see if it gets fixed before moving all the fake timers to epoch.


.github/workflows/build-and-deploy.yml
```diff
@@ -10,7 +10,6 @@ jobs:
         uses: actions/checkout@v2
         with:
           registry-url: 'https://npm.pkg.github.com'
-      - run: echo "registry=https://registry.npmjs.org/" > .npmrc
       - run: echo "@internxt:registry=https://npm.pkg.github.com" >> .npmrc
```
sg-gs: Remove all the tricks, not only the first line (comments related to this step that are behind, etc)

sg-gs (approved):

PR #1039, opened 2026-04-16
[_] fix: add limit 1 to backup queries

apsantiso:
## What
Some queries were missing a LIMIT 1, causing us to fetch all folders/files inside a backup just for a simple "isEmpty" check.
## Changes

- Removed circular dependency between folderUsecases and fileUsecases.
- Fetch only 1 folder when checking the availability of a backup name.
- Fetch only 1 folder and 1 file when checking whether the device has any backups.


sg-gs (approved):
Good catch @apsantiso

PR #1047, opened 2026-04-28
[PB-6327] feat(file-size-limit): limit file size on create and replace

apsantiso:
This prepares the environment for file size limit. 
### Changes
- Created migration for file size limits
- Added file size limit check for not empty files

### DIFF
--- a/migrations/20260428065637-add-max-upload-file-size-limit.js
+++ b/migrations/20260428065637-add-max-upload-file-size-limit.js
@@ -1,0 +1,63 @@
+'use strict';
+
+const { v4 } = require('uuid');
+
+const LIMIT_LABEL = 'max-upload-file-size';
+
+const MB = 1024 * 1024;
+const GB = 1024 * MB;
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.bulkInsert('limits', [
+      {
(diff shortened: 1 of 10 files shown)


sg-gs (commented):
We should expire the cache if a tier change occurs. Is there any point in these changes where that is handled? 

This needs to be done if we are going to rely on the cache when possible, otherwise weird situations like not being able to do something after upgrading or allowing something temporarily after downgrading could happen.

apsantiso:
@sg-gs The cache is for the tier (globally), not the user. If we increase limit for a plan someday, we can just wait 10 minutes for the TTL to expire. 

I am thinking about adding a cache for the user's tier, overriden limits or whatever setting that we can just remove/upgrade when the user has changed the plan like we do with the limit, but that would be a nice-to-have that I preferred not to handle in this ticket.

sg-gs (approved):
Running migration @apsantiso

sg-gs:
Migration run @apsantiso 

```
$ cross-env NODE_ENV=production sequelize db:migrate

Sequelize CLI [Node: 22.16.0, CLI: 6.6.3, ORM: 6.37.7]

== 20260428065637-add-max-upload-file-size-limit: migrating =======
== 20260428065637-add-max-upload-file-size-limit: migrated (0.077s)

Done in 1.01s.
```

PR #1069, opened 2026-05-27
[PB-6412] feat: added photos backups as folders

apsantiso:
Added photos backups.  The idea is to follow the same principles as deviceAsFolder.

I wanted to avoid creating a new field in the users table, but as we do not have other ways to save buckets (for the time being) and to introduce this without too much changes, it is more straightforward to just add the bucket to that table like we do with backups.

## Changes
- Added migration for photos_bucket in users table
- Added migration for indexing bucket again
- Added photos backup creation endpoint
- Added photos backup modification endpoint
- Added photos backup listing endpoint
- Added photos backup get endpoint


migrations/20260527141826-recreate-index-folders-bucket.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(
+      `CREATE INDEX CONCURRENTLY IF NOT EXISTS bucket_index ON folders (bucket) WHERE deleted = false AND removed = false`,
```
apsantiso: The index was removed from prod, but we look by bucket when fetching any device as folder.

migrations/20260527141812-add-photos-bucket-to-users.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('users', 'photos_bucket', {
+      type: Sequelize.STRING,
+      allowNull: true,
```
apsantiso, line 7: It is nullable so it should not lock the table for more time than just the required to add the column.

migrations/20260527141826-recreate-index-folders-bucket.js
sg-gs: Does the device have parent folder ? If not, add the parent_id NULL condition to the index, so we avoid indexing the whole folders-existing part just for a minority of records that act as 'devices'

src/modules/backups/backup.controller.ts
```diff
@@ -236,6 +236,64 @@ export class BackupController {
     return this.backupUseCases.getDevicesAsFolder(user);
   }
 
+  @Post('/photos/devices')
```
sg-gs: I would create a photos controller, although the service is closely related to backups

sg-gs (approved):
Running migration on Monday

sg-gs:
Running migration @apsantiso

sg-gs:
Done @apsantiso 

```

== 20260527141812-add-photos-bucket-to-users: migrating =======
== 20260527141812-add-photos-bucket-to-users: migrated (0.066s)

== 20260527141826-recreate-index-folders-bucket: migrating =======
== 20260527141826-recreate-index-folders-bucket: migrated (471.933s)
```

PR #1073, opened 2026-06-15
[_]: feat/improve-path-limit

larryrider:
It changes the path validation from depth-based to length-based

## Problem
The file/folder metadata-by-path endpoints currently reject any path with a depth greater than 20 segments (getPathDepth(path) > 20). This is overly restrictive, it blocks legitimate deep directory trees that are well within OS limits.
While most users dont hit this limit, we have some reports from users whose paths exceed 20 segments. Forcing them to restructure their folder hierarchy is a poor experience, and the number of affected users is small enough that there is no meaningful performance concern in allowing deeper paths.

## Solution
Replace the depth check with a path length check (path.length > 1024), matching the behaviour of macOS and aligning more closely with common OS path limits:

macOS -> 1024 characters
Linux -> 4096 characters
Windows -> 260 characters (default) / ~32,767 characters (long paths)

Using 1024 characters as the limit provides a generous safety margin below Linux's 4096 limit while allowing deeply nested but legitimate paths that the old 20-segment cap would have rejected. It also matches macOS's documented limit, which is a common reference for some cross-platform tools.
Changes
- src/modules/file/file.controller.ts -> Changed validation from getPathDepth(filePath) > 20 to filePath.length > 1024; removed unused getPathDepth import.
- src/modules/folder/folder.controller.ts -> Same change for folder metadata lookup.
- src/modules/file/file.controller.spec.ts -> Updated test to generate path > 1024 characters instead of 21 segments.
- src/modules/folder/folder.controller.spec.ts -> Same for folder test.
- src/lib/path.ts -> The getPathDepth helper remains available for other uses but is no longer called from the controllers.

### DIFF
--- a/src/modules/file/file.controller.spec.ts
+++ b/src/modules/file/file.controller.spec.ts
@@ -191,11 +191,8 @@ describe('FileController', () => {
       );
     });
 
-    it('When get file metadata by path is requested with a path deep > 20, then it should throw an error', () => {
-      const longPath =
-        '/' +
-        Array.from({ length: 21 }, (_, i) => `folder${i}`).join('/') +
-        '/file.test';
+    it('When get file metadata by path is requested with a path length > 1024, then it should throw an error', () => {
+      const longPath = '/' + 'a'.repeat(1025);
 
       expect(
         fileController.getFileMetaByPath(userMocked, longPath),
--- a/src/modules/file/file.controller.ts
+++ b/src/modules/file/file.controller.ts
@@ -468,7 +468,11 @@ export class FileController {
     @UserDecorator() user: User,
     @Query('path') filePath: string,
   ): Promise<FileDto> {
-    if (!filePath || filePath.length === 0 || !filePath.includes('/')) {
+    if (
+      typeof filePath !== 'string' ||
+      filePath.length === 0 ||
+      !filePath.includes('/')
+    ) {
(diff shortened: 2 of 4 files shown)


sg-gs:
Which is the database impact? The depth of something is also important for operating the service as we use a linked list  to represent the folder hierarchy and this is one of the biggest trade-offs. Could you provide a explain analyze of let's say, as much depth as possible within the new limits in the worst case (each folder's name has only one letter) of the involved queries to operate this path-based feature so we can see the performance penalty? @larryrider

larryrider:
> Which is the database impact? The depth of something is also important for operating the service as we use a linked list to represent the folder hierarchy and this is one of the biggest trade-offs. Could you provide a explain analyze of let's say, as much depth as possible within the new limits in the worst case (each folder's name has only one letter) of the involved queries to operate this path-based feature so we can see the performance penalty? @larryrider

Hey @sg-gs! This is the result of the explain analyze of the worst case scenario (512 subfolders) 
image


This is the one from the query itself (just for one folder):
image


We should multiply that cost by 512 (which gives around 1k)

sg-gs (approved):
LGTM.

apsantiso (approved):
@larryrider LGTM, let's see how it behaves when this starts to get used. We'll have a better idea then.

PR #1082, opened 2026-06-17
[_] Fix file sharing code

TamaraFinogina:
File sharing was calling key generation function directly. Switched to the common function

### DIFF
--- a/src/modules/sharing/sharing.service.ts
+++ b/src/modules/sharing/sharing.service.ts
@@ -246,12 +246,13 @@ export class SharingService {
           clientVersion: '1.0.0',
         },
       });
-      const fileInfo = await network.getFileInfo(item.bucket, item.fileId);
 
-      const encryptionKey = await Environment.utils.generateFileKey(
-        aes.decrypt(sharing.encryptionKey, code),
-        item.bucket,
-        Buffer.from(fileInfo.index, 'hex'),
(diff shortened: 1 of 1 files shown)


src/modules/sharing/sharing.service.ts
```diff
@@ -246,20 +246,21 @@ export class SharingService {
           clientVersion: '1.0.0',
         },
       });
-      const fileInfo = await network.getFileInfo(item.bucket, item.fileId);
```
TamaraFinogina, line 249: @apsantiso It was marked as not used, not sure why

TamaraFinogina: never mind, see it

src/modules/sharing/sharing.service.ts
```diff
@@ -246,20 +246,21 @@ export class SharingService {
...
+        item,
+        sharing.encryptionKey,
+        code,
+        network,
+        sharing.encryptionAlgorithm === NEW_SHARING_VERSION,
       );
       response['itemToken'] = await network.createFileToken(
         item.bucket,
         item.fileId,
         'PULL',
```
TamaraFinogina, line 260: Image

should be the same result

jzunigax2 (approved):

apsantiso (approved):

apsantiso:
@TamaraFinogina Feel free to merge

sg-gs (approved):

PR #1083, opened 2026-06-17
[_] fix: add keys check

apsantiso:
## What

Input keys format / bytes validation.

- `IsOpenPgpPublicKey()` — Ecc `publicKey`
- `IsKyberPublicKey()` — Kyber `publicKey`
- `IsEncryptedKeyOfSize(exactPayloadBytes?)` — both `privateKey` fields

## Why

Before this, `publicKey`/`privateKey` only had `@IsString() @IsNotEmpty()`, any string passed.

## How

**publicKey**: real checks against known constraints.
- Ecc: `openpgp.readKey()` must parse + `!key.isPrivate()`. Real structural parse, not heuristic.
- Kyber512: must decode to exactly 800 bytes (confirmed via `kem.publicKeyBytes` from `@dashlane/pqc-kem-kyber512-node`). We do not use the library here because NestJS fails getting ESM only modules.

**privateKey**: both are AES-256-GCM blobs (`salt(64)+iv(16)+tag(16)+payload`), base64. Server never sees plaintext key:
- Kyber: exact size works. Client encrypts the raw key's base64 string; raw key is always 1632 bytes → base64 = 2176 bytes → full blob = 96 (AES headers) +2176 = **2272 bytes exact**. Verified with real keypair + real `AesService.encrypt()`.
- Ecc: varies between 903 and 907. Also, according to the openpgp version the key is heavier (there's a difference of ~200 bytes between the keys created with openpgp 5 and 6).
  - ℹ️  I've omitted this test as this repository openpgp version is still 5 and I was not able to create good e2e tests to secure a smooth deployment. We can update in a later PR if required.

### DIFF
--- a/src/externals/crypto/decorators/is-encrypted-mnemonic.decorator.spec.ts
+++ b/src/externals/crypto/decorators/is-encrypted-mnemonic.decorator.spec.ts
@@ -35,18 +35,6 @@ describe('IsEncryptedMnemonic', () => {
     });
   });
 
-  it('When encrypted 12-word mnemonic is provided, then validation error returned', async () => {
-    const mnemonic = generateMnemonic(128);
-    const encrypted = encryptMnemonicForTest(mnemonic, 'password');
-
-    const errors = await validate(buildDto(encrypted));
-
-    expect(errors.length).toBeGreaterThan(0);
-    expect(errors[0].constraints).toMatchObject({
-      minLength: 'Invalid encrypted mnemonic.',
-    });
-  });
-
   it('When mnemonic is too long, then validation error returned', async () => {
     const errors = await validate(buildDto('a'.repeat(481)));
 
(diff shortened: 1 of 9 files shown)


src/externals/crypto/decorators/is-encrypted-mnemonic.decorator.spec.ts
```diff
@@ -35,18 +35,6 @@ describe('IsEncryptedMnemonic', () => {
     });
   });
 
-  it('When encrypted 12-word mnemonic is provided, then validation error returned', async () => {
```
sg-gs, line 38: Why?

apsantiso: It's a flaky test. 12 words mnemonics can reach the 24 words minimum threshold sometimes when long words are used so adding randomness here makes the test fails sometimes.

sg-gs (approved):

TamaraFinogina:
I'm a bit afraid that base64 can sometimes give us a bit longer/shorter string due to encoding. But key sizes are corect

TamaraFinogina (approved):

PR #1102, opened 2026-07-13
[PB-6596] feat: favorites table, model, domain, repository and use cases

victor-ferro:
## What

Foundation for the favorites feature (no API surface yet — endpoints come in the follow-up PR):

- **[PB-6596]** `5739071f` — `favorites` table migration (unique on `user_id, item_id, item_type`, FK to `users.uuid` with `ON DELETE CASCADE`), `FavoriteModel`, `Favorite` domain, `isFavorite` field on file/folder domains and DTOs, and the `favorites` association on `FileModel`/`FolderModel`.
- **[PB-6605]** `94f35b45` — `SequelizeFavoriteRepository` (create/delete/bulkDelete/existsForUser), `FavoriteUseCases` (mark/unmark with existence + ownership checks, bulk removal), `FavoriteModule` registered in `app.module`.

Nothing consumes the module yet, so merging this changes no existing behavior.

## Testing

Unit tests colocated per layer (domain, repository, use cases). Each commit was verified in isolation: `npx jest src/modules/favorite` — 3 suites, 22 tests passing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[PB-6596]: https://inxt.atlassian.net/browse/PB-6596?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PB-6605]: https://inxt.atlassian.net/browse/PB-6605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1103, opened 2026-07-13
[PB-6606] feat: favorites endpoints and isFavorite flag in listings

victor-ferro:
## What

API surface for favorites, on top of #1102:

- **[PB-6606]** `194e210e` — `GET /files/favorites` and `GET /folders/favorites` (cursor pagination, `updatedAt` filter, sort), plus `isFavorite` annotation in existing file/folder listings via an opt-in `favoriteUserUuid` param that adds a per-user join against `favorites` (`separate: true`, so existing queries and their limit/offset are untouched when the param is absent).
- **[PB-6607]** `9de31428` — `PUT /:uuid/favorite` / `DELETE /:uuid/favorite` on files and folders (ownership enforced in `FavoriteUseCases`), and favorites cleanup alongside the existing sharings/thumbnails cleanup on permanent deletes.

## Testing

Unit tests per layer (DTO validation, repositories, use cases, controllers). Verified per commit: 8 suites / 491 tests after PB-6606, and the full suite (136 suites / 2780 tests) after PB-6607.

## Notes for review

- Depends on #1102 — this PR targets `PB-6596-favorites-base`; once #1102 merges, GitHub will retarget it to `master`.
- Commits are reviewable independently if you prefer smaller chunks (GET endpoints vs mark/unmark + cleanup).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[PB-6606]: https://inxt.atlassian.net/browse/PB-6606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PB-6607]: https://inxt.atlassian.net/browse/PB-6607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/file/file.controller.ts
```diff
@@ -505,6 +509,24 @@ export class FileController {
...
+  @ApiOkResponse({ isArray: true, type: FileDto })
+  async getFavoriteFiles(
+    @UserDecorator() user: User,
+    @Query() queryParams: GetFavoriteFilesDto,
+  ): Promise<FileDto[]> {
+    const { limit, offset, sort, order, updatedAt } = queryParams;
+
+    return this.fileUseCases.getFavoriteFiles(
+      user,
+      new Date(updatedAt || 1),
```
apsantiso: Is there a case where you need to filter all the files updated after a specific time? If not, then you can just remove this and the reference from all the favorites usecases. If you need it then leave a comment an feel free to mark this comment as resolved 👍🏻

src/modules/folder/folder.controller.ts
```diff
@@ -818,6 +832,34 @@ export class FolderController {
...
+    );
+
+    return folders.map((f) => {
+      if (!f.plainName) {
+        f.plainName = this.folderUseCases.decryptFolderName(f).plainName;
+      }
+
+      return { ...f, status: f.getFolderStatus() };
+    });
+  }
```
apsantiso: As this a non-essential feature, what about just creating another controller in the favorites module? So instead of having 3 new endpoints here (6 in total with files), you can just have 3 centralized endpoints?

src/modules/file/file.controller.ts
```diff
@@ -505,6 +509,24 @@ export class FileController {
     }
   }
 
+  @Get('/favorites')
```
apsantiso: As this a non-essential feature, what about just creating another controller in the favorites module? So instead of having 3 new endpoints here (6 in total with folders), you can just have 3 centralized endpoints?

src/modules/folder/folder.controller.ts
```diff
@@ -818,6 +832,34 @@ export class FolderController {
...
+  async getFavoriteFolders(
+    @UserDecorator() user: User,
+    @Query() query: GetFavoriteFoldersDto,
+  ): Promise<FolderDto[]> {
+    const sort: SortParamsFolder | undefined =
+      query.sort && query.order ? [[query.sort, query.order]] : undefined;
+
+    const folders = await this.folderUseCases.getFavoriteFolders(
+      user,
+      new Date(query.updatedAt || 1),
```
apsantiso: Is there a case where you need to filter all the folders updated after a specific time? If not, then you can just remove this, otherwise, you can leave a comment a resolve this comment.

src/modules/folder/folder.usecase.ts
```diff
@@ -1028,6 +1063,10 @@ export class FolderUseCases {
     await this.folderRepository.deleteByUser(user, folders);
   }
 
+  // Unlike FileUseCases.deleteFilePermanently, this doesn't clean up sharings
+  // or favorites rows referencing these folders (pre-existing gap for sharings,
+  // not introduced by favorites). Harmless for listings, since both queries
+  // already filter by deleted/removed, but leaves orphaned rows behind.
```
apsantiso: Remove this comment or change it to a short `FIXME:` comment so it gets marked as tech debt

apsantiso (changes requested):

src/modules/file/file.controller.ts
victor-ferro: Delete!

src/modules/folder/folder.controller.ts
victor-ferro: Right! I'll replicate the trash functionality

src/modules/folder/folder.usecase.ts
victor-ferro: Removed the comment and implemented the cleanup instead. DELETE /folders/:uuid now removes the folder's favorites rows alongside the deletion

apsantiso (approved):

PR #1106, opened 2026-07-22
[PB-6630] feat: add fuzzy search filters query contract

victor-ferro:
## What

First of two stacked PRs for **[PB-6630] advanced search filters** in the fuzzy search endpoints (`GET /fuzzy/:search` and `GET /workspaces/:workspaceId/fuzzy/:search`). This one introduces the query contract only — nothing consumes it yet, so there is **no behavior change**. The second PR wires it into repository, use cases and controllers.

- **`FuzzySearchQueryDto`**: validated query params for the search endpoints — `offset`, `type` (repeatable file category, single value auto-wrapped into an array), `minSize`/`maxSize` (bytes, non-negative integers, cross-validated so `maxSize >= minSize`), `modifiedAfter`/`modifiedBefore` (ISO 8601). Follows the existing class-validator + global `ValidationPipe` pattern.
- **`FileCategory` + `FILE_CATEGORY_EXTENSIONS`**: category → extensions mapping (audio, code, image, pdf, video…), mirroring drive-web's `file-types.ts` so the frontend can send categories instead of raw extension lists. `folder` is a pseudo-category handled at query-build time. Marked with a TODO to extract into a shared package later.
- **`resolveCategoriesToExtensions`**: flattens categories into a deduplicated extension list (some extensions live in more than one category).

## Testing

- `fuzzy-search-query.dto.spec.ts`: 12 unit tests covering transforms (string→number, single value→array), invalid categories, negative/non-numeric sizes, `maxSize < minSize` rejection and invalid dates.

[PB-6630]: https://inxt.atlassian.net/browse/PB-6630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apsantiso (dismissed):

apsantiso (approved):

src/modules/fuzzy-search/file-categories.ts
```diff
@@ -0,0 +1,231 @@
...
+    'f4v',
+    'f4p',
+    'f4a',
+    'f4b',
+  ],
+  [FileCategory.Word]: ['doc', 'docx'],
+  [FileCategory.Xls]: ['xls', 'xlsx'],
+  [FileCategory.Xml]: ['xml', 'xsl', 'xsd', 'svg'],
+  [FileCategory.Zip]: ['zip', 'zipx'],
+};
```
apsantiso: Isn't better that this lives in the frontend and the backend just accepts whatever type you send (setting a max length)? You won't have duplicated logic that way and the backend stays agnostic

apsantiso (changes requested):

src/modules/fuzzy-search/file-categories.ts
apsantiso: Nothing would break if you pass something that isn't a file extension to the fuzzy-search, so it's ok to be flexible here 🙂

If you want to add some validations, add a maxLength or something simple to prevent funny users from sending an entire paragraph haha.

EDIT: it's just an idea, as I'm not sure how long the URL could get if you select all the categories. Please do not take it as a must

victor-ferro: Yes, this could be an option. we actually pulled this list from drive-web's file-types.ts. I didn't do it that way from the start for three reasons:
- Is drive-web going to be the only client of this endpoint? Because if mobile or any other client calls it now or in the future, the list would also have to be duplicated in those clients.

- The size issue you mention. Audio and code alone have 90 sub-types. If categories with many types are selected, we could end up with a really long query.

I think the perfect solution would be to create a package to store these types and keep everything always in sync, but of course that would be a separate task. If you want me to keep it only in the frontend and remove it from here, confirm it and I'll do it

apsantiso: No other clients use this for the time being

victor-ferro: Ah okay, should I change it then?

apsantiso: Yes please, to keep things simple, leave it in the frontend, if you need to centralize it, then add it to the SDK or wherever is required

victor-ferro: Done!

src/main.ts
```diff
@@ -51,7 +53,9 @@ async function bootstrap() {
 
   app.set('trust proxy', enableTrustProxy);
   // Express v5 changed this to 'simple' by default, we need the same behavior as v4
-  app.set('query parser', 'extended');
+  app.set('query parser', (query: string) =>
+    parse(query, { arrayLimit: MAX_TYPE_FILTERS }),
+  );
```
apsantiso: Let's use a post request instead of a GET so you do not need to do this, as it increases the limit for all the EPs. It's OK not to follow REST in this cases, don't worry

victor-ferro: Done!

apsantiso (approved):

PR #1107, opened 2026-07-22
[PB-6630] feat: apply search filters in fuzzy search endpoints

victor-ferro:
## What

Second of two stacked PRs for **[PB-6630] advanced search filters**. Building on the query contract from #1106, this wires the filters (`type`, `minSize`/`maxSize`, `modifiedAfter`/`modifiedBefore`) into both fuzzy search endpoints, combined with AND logic and applied **before** the `LIMIT 10` — so filtering happens over the whole dataset instead of the returned page.

- **`look-up.repository.ts`**: a single `buildFilterClauses` builds the SQL fragments and replacements shared by `search` and `workspaceSearch` (values always parameterized). Extension filtering via `LOWER(f."type") IN (:extensions)`, size via `f."size"`, dates via `modification_time` on both tables (already indexed — no new migrations). Size filters exclude the folders branch of the `UNION ALL` (folders have no size); `type=folder` toggles it on. Contradictory combinations (e.g. `folder` + `minSize`) return `[]` without querying.
- **`fuzzy-search.usecase.ts`**: translates the validated `FuzzySearchQueryDto` into repository filters — resolves categories into extensions, maps the `folder` pseudo-category into item types.
- **Controllers** (`fuzzy-search` and `workspaces`): swap the loose `offset` pipes for the shared `@Query() FuzzySearchQueryDto`, keeping both endpoints' contract identical.

## Testing

- `look-up.repository.spec.ts` (new): 10 tests asserting the generated SQL fragments/replacements per filter, the contradictory-filters short-circuit, and that both search variants share the same clause builder.
- `fuzzy-search.usecase.spec.ts`: 6 new tests for the DTO→filters translation (category resolution, folder handling, extension dedup, size/date passthrough).
- Existing workspaces/controller specs updated to the new signature.

## Related Pull Requests

- #1106 — `[PB-6630] feat: add fuzzy search filters query contract` (base of this PR; merge it first, then this PR gets re-targeted to `master`)

[PB-6630]: https://inxt.atlassian.net/browse/PB-6630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PB-6630]: https://inxt.atlassian.net/browse/PB-6630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/fuzzy-search/look-up.repository.ts
```diff
@@ -16,21 +16,82 @@ type LookUpResult = Array<{
...
+  replacements: Record<string, unknown>;
+}
+
+function buildFilterClauses(filters: FuzzySearchFilters): FilterClauses {
+  const includeFiles = !filters.itemTypes || filters.itemTypes.includes('file');
+  const includeFolders =
+    (!filters.itemTypes || filters.itemTypes.includes('folder')) &&
+    filters.minSize === undefined &&
+    filters.maxSize === undefined;
+
```
jzunigax2, line 72: I understand we could search by either files, folders or both right? 

So if folders were explictly included in the search would it make sense to disable their search if file size is also sent? Either way makes sense to me, their inclusion or removal from the filter just flagging it.

However I do see a possible issue. A minSize 0 would trigger this folder exclusion maybe you could harden this logic a bit?

victor-ferro: Good catch. Fixed by deriving both the folder exclusion and the SQL clause from the same check, minSize > 0, so a zero no longer counts as a size filter. maxSize: 0 still excludes folders, since files of exactly 0 bytes is a real filter. Two tests added, one per zero

jzunigax2: great, that makes sense

jzunigax2 (dismissed):
lgtm, though I guess this should not be merged yet as it breaks web search

victor-ferro:
@jzunigax2 You're right. What if we bring the GET back and keep both for now, then drop it in a cleanup PR once the SDK and the new front are out? That way search never breaks in between and we don't have to line up the three deploys

jzunigax2:
> @jzunigax2 You're right. What if we bring the GET back and keep both for now, then drop it in a cleanup PR once the SDK and the new front are out? That way search never breaks in between and we don't have to line up the three deploys

alright, just be careful to bring back the original logic then. meaning, leave the current methods as they were before and move the new logic onto new ones. Sonar will surely cry about code duplication but we'll drop the old ones later on

victor-ferro:
The GET endpoints are back with their original logic untouched — the diff against master is additive only, no lines removed. @jzunigax2

src/modules/fuzzy-search/look-up.repository.ts
```diff
@@ -204,4 +279,172 @@ export class SequelizeLookUpRepository implements LookUpRepository {
...
+          CASE WHEN f."plain_name" = :partialName THEN 1 ELSE 0 END            AS "exactMatch",
+          f."type"        AS "file.type",
+          f."id"          AS "file.id",
+          f."size"        AS "file.size",
+          f."bucket"      AS "file.bucket",
+          f."file_id"     AS "file.fileId",
+          f."plain_name"  AS "file.plainName",
+          NULL            AS "folder.id"
+      FROM files f
+      WHERE f."user_id" = (SELECT id FROM users WHERE uuid = :userUuid)
```
jzunigax2: great, now my only concern is that now the base query gets repeated a bunch I know I said not to worry about code duplication but I was referring to maybe similarities between the old and new code. maybe the base query could get extracted to a const or injected via a function?

jzunigax2 (approved):

PR #1108, opened 2026-07-24
[MKT_949]:feat/integrate turnstile

jaaaaavier:
This PR tries to Add support for Cloudflare Turnstile as an alternative CAPTCHA provider alongside the existing Google reCAPTCHA v3. The provider is selectable at runtime via a new CAPTCHA_PROVIDER environment variable, so we can switch between reCAPTCHA and Turnstile without code changes.

Changes:
- captcha.service.ts: Refactored CaptchaService to route verification based on the configured provider. The old inline reCAPTCHA logic is now extracted into a private verifyRecaptcha() method, and a new verifyTurnstile() method handles ------Turnstile's siteverify endpoint. Shared error handling/logging remains in the public verify() method.
configuration.ts: Added captcha.provider (defaults to recaptcha), captcha.turnstileUrl (defaults to Cloudflare's siteverify endpoint), and secrets.turnstileCaptcha.
-.env.template: Added TURNSTILE_ENDPOINT, TURNSTILE_SECRET, and CAPTCHA_PROVIDER.
-captcha.service.spec.ts: New test suite covering both providers, including success, failure, below-threshold (reCAPTCHA), and error scenarios.

### Notes
We will need new Environment variables

### DIFF
--- a/src/externals/captcha/captcha.service.spec.ts
+++ b/src/externals/captcha/captcha.service.spec.ts
@@ -1,0 +1,128 @@
+import axios from 'axios';
+import { CaptchaService } from './captcha.service';
+
+jest.mock('axios');
+
+const mockedAxios = axios as jest.Mocked<typeof axios>;
+
+describe('CaptchaService', () => {
+  let service: CaptchaService;
+  const originalEnv = process.env;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    service = new CaptchaService();
+    process.env = { ...originalEnv, NODE_ENV: 'production' };
+  });
+
+  afterAll(() => {
+    process.env = originalEnv;
+  });
+
+  it('When not in production, then it skips verification and returns true', async () => {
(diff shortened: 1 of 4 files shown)


src/externals/captcha/captcha.service.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    jest.clearAllMocks();
+    service = new CaptchaService();
+    process.env = { ...originalEnv, NODE_ENV: 'production' };
+  });
+
+  afterAll(() => {
+    process.env = originalEnv;
+  });
+
+  it('When not in production, then it skips verification and returns true', async () => {
```
xabg2, line 22: Use test instead of it.

apsantiso:
Added @xabg2 as reviewer as I do not have too much idea what is this guard used for so I do not know if there's anything that could break.

apsantiso:
Open it again if you need it

PR #1111, opened 2026-07-30
[PB-6596] fix: favorites fixes from QA review — trash/restore behaviour and listing completeness

victor-ferro:
## What

Fixes for the favorites QA review, on top of the merged favorites feature (#1103):

- **[PB-6596]** `823f6545` — **Favorites survive trash and restore.** Favorite rows are no longer deleted when items are moved to trash, so restoring an item also restores its favorite status. The favorites listing now excludes items whose ancestor folders are trashed or removed (recursive ancestors check in `findAllCursorFavorites`, same depth guard as `calculateFolderSize`). Rows are actually cleaned up on permanent deletion (single item, empty trash and backups), with an orphan sweep after trash deletion finishes.
- **[PB-6596]** `203a6f49` — **Favorite files listing includes thumbnails and sharings**, reusing the same `separate: true` includes as the Drive listing, so the Favorites page renders file thumbnails and the shared badge exactly like Drive (*"Same list component as in Drive"*).

## Testing

- Unit: 437 tests green across the 10 touched suites, re-run on top of current master (after `chore: removefile Id`).
- SQL semantics validated against a live Postgres with synthetic trees (inside rolled-back transactions): favorites under trashed/removed ancestors are excluded at any depth; alive trees are unaffected.
- Manually verified end-to-end with drive-web (paired fixes on `PB-5666-favorites-empty-state`): trashing a folder with nested favorites removes them from the Favorites page instantly, restoring brings them all back, and a favorited PDF renders its thumbnail plus the shared badge.

## Notes for review

- Keeping favorite rows through trash is intentional — it is what makes restore recover them; permanent deletion is where rows are removed. Rows belonging to items shared by other users are out of scope here (favorites-on-shared-items is pending a team decision and lives in a separate stash).

[PB-6596]: https://inxt.atlassian.net/browse/PB-6596

### DIFF
--- a/src/modules/favorite/favorite.repository.ts
+++ b/src/modules/favorite/favorite.repository.ts
@@ -114,18 +114,22 @@ export class SequelizeFavoriteRepository implements FavoriteRepository {
     userId: Favorite['userId'],
     folderUuids: string[],
   ): Promise<void> {
+    const MAX_DEPTH = 10000;
+
     await this.favoriteModel.sequelize.query(
       `
       WITH RECURSIVE subtree AS (
-        SELECT fl1.uuid
+        SELECT fl1.uuid, 1 AS depth, ARRAY[fl1.uuid] AS path
         FROM folders fl1
         WHERE fl1.uuid IN (:folderUuids)
 
-        UNION
+        UNION ALL
 
-        SELECT fl2.uuid
+        SELECT fl2.uuid, subtree.depth + 1, subtree.path || fl2.uuid
         FROM folders fl2
         INNER JOIN subtree ON fl2.parent_uuid = subtree.uuid
+        WHERE subtree.depth < :maxDepth
+          AND NOT fl2.uuid = ANY(subtree.path)
       )
       DELETE FROM favorites
       WHERE user_id = :userId
@@ -144,7 +148,7 @@ export class SequelizeFavoriteRepository implements FavoriteRepository {
         )
       `,
       {
-        replacements: { userId, folderUuids },
+        replacements: { userId, folderUuids, maxDepth: MAX_DEPTH },
         type: QueryTypes.DELETE,
       },
     );
--- a/src/modules/favorite/favorite.repository.ts
+++ b/src/modules/favorite/favorite.repository.ts
@@ -116,32 +116,31 @@ export class SequelizeFavoriteRepository implements FavoriteRepository {
   ): Promise<void> {
     await this.favoriteModel.sequelize.query(
       `
+      WITH RECURSIVE subtree AS (
+        SELECT fl1.uuid
+        FROM folders fl1
+        WHERE fl1.uuid IN (:folderUuids)
+
+        UNION
+
+        SELECT fl2.uuid
+        FROM folders fl2
+        INNER JOIN subtree ON fl2.parent_uuid = subtree.uuid
+      )
       DELETE FROM favorites
       WHERE user_id = :userId
-        AND EXISTS (
-          WITH RECURSIVE ancestors AS (
-            SELECT fl1.uuid, fl1.parent_uuid, 1 AS depth
-            FROM folders fl1
-            WHERE fl1.uuid = CASE
-              WHEN favorites.item_type = 'file' THEN (
-                SELECT files.folder_uuid FROM files
-                WHERE files.uuid = favorites.item_id
-              )
-              ELSE (
-                SELECT folders.parent_uuid FROM folders
-                WHERE folders.uuid = favorites.item_id
-              )
-            END
-
-            UNION ALL
-
-            SELECT fl2.uuid, fl2.parent_uuid, ancestors.depth + 1
-            FROM folders fl2
-            INNER JOIN ancestors ON fl2.uuid = ancestors.parent_uuid
-            WHERE ancestors.depth < 100000
-          )
-          SELECT 1 FROM ancestors
-          WHERE ancestors.uuid IN (:folderUuids)
+        AND (
+          (favorites.item_type = 'folder' AND EXISTS (
+            SELECT 1 FROM folders
+            WHERE folders.uuid = favorites.item_id
+              AND folders.parent_uuid IN (SELECT uuid FROM subtree)
+          ))
+          OR
+          (favorites.item_type = 'file' AND EXISTS (
+            SELECT 1 FROM files
+            WHERE files.uuid = favorites.item_id
+              AND files.folder_uuid IN (SELECT uuid FROM subtree)
+          ))
         )
       `,
       {
--- a/src/modules/favorite/favorite.repository.ts
+++ b/src/modules/favorite/favorite.repository.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@nestjs/common';
 import { InjectModel } from '@nestjs/sequelize';
-import { Op } from 'sequelize';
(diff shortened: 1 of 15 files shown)


src/modules/folder/folder.repository.ts
```diff
@@ -468,6 +469,25 @@ export class SequelizeFolderRepository implements FolderRepository {
...
+        UNION ALL
+
+        SELECT fl2.uuid, fl2.parent_uuid, fl2.deleted, fl2.removed, ancestors.depth + 1
+        FROM folders fl2
+        INNER JOIN ancestors ON fl2.uuid = ancestors.parent_uuid
+        WHERE ancestors.depth < 100000
+      )
+      SELECT 1 FROM ancestors
+      WHERE ancestors.deleted = TRUE OR ancestors.removed = TRUE
+    )`);
```
apsantiso: Beware of this as it will lead to **_really_** bad performance if the impact isn't measured correctly.

This might work OK locally, but in prod **_maybe_** we can't afford to navigate the folder tree for all the favorited items in the list without compromising a lot of performance

Have you performed an explain analyze on this query for a user with 100k files/folders? It would be good for measuring the impact on prod.

It's better that you push back on the behaviors that would affect the app's performance. If persisting the favorite state when an item is being trashed is too much trouble, push back. The product team should understand

victor-ferro: Measured it. Throwaway DB with a 103k-item account (21,845 folders + 81,920 files, depth 7, no trashed ancestors = worst case), running the real repository code and capturing the SQL Sequelize emits. Median of 5 runs, warm cache.

Files listing, limit 50, with jit on (the prod default):

| favorites | current | without `NOT EXISTS` | factor |
|---|---|---|---|
| 500 | 38.5 ms | 1.8 ms | 21x |
| 5,000 | 384.7 ms | 18.3 ms | 21x |
| 10,000 | 448.0 ms | 26.8 ms | 17x |

So yes, the impact is real and larger than expected. Two things worth keeping in mind though:

**The scaling axis isn't the number of items in the account.** The `NOT EXISTS` runs after the join with `favorites`, so `loops` is the number of favorites, not the number of items. A user with 100k files and 50 favor100,000. Cost is`favorites × depth`, so we can bound the worst case with a per-user favorites limit.

**About ~70% of that cost isn't query work, it's compilation.** Theplanner estimates TE when the reality is 8 rows per iteration, and that overestimation crosses the       threshold that triit off, the 10,000case drops from 448 ms to 152 ms, and the 500 case from 38.5 ms to 13.3 ms. That takeo ~7x, so it'd beworth looking into whether we can keep it from triggering: either by reformulating the query so it doesn't inflate the estimate, or by disabling jit fe used correctly onevery hop, no avoidable seq scans.

victor-ferro:
**Favorites are cleaned up on delete instead of filtered on read**

Trashing an item removes its favorite again, and `deleteInsideFoldersByUser` now also removes the favorites of everything inside a deleted folder, at any depth, in a single recursive-CTE `DELETE`.

This replaces the previous approach, where favorites survived trash and the listing had to hide them with a correlated `NOT EXISTS (WITH RECURSIVE ancestors …)` evaluated per row **on every read** of the Favorites view — non-indexable, so `ORDER BY … LIMIT` had to filter row by row.

**Cost:** the ancestor walk moves from read-time to write-time — once per delete operation instead of once per listing, and only when a folder is actually deleted. The delete goes bottom-up from the user's favorites, so it is independent of the deleted subtree's size: roughly `favorites × depth` index probes (200 favorites, depth 8 → ~1.8k probes, single-digit ms), with recursion capped at 100k levels. Trade-off: restoring from trash no longer restores favorite status.

src/modules/favorite/favorite.repository.ts
```diff
@@ -81,6 +86,71 @@ export class SequelizeFavoriteRepository implements FavoriteRepository {
...
+
+            SELECT fl2.uuid, fl2.parent_uuid, ancestors.depth + 1
+            FROM folders fl2
+            INNER JOIN ancestors ON fl2.uuid = ancestors.parent_uuid
+            WHERE ancestors.depth < 100000
+          )
+          SELECT 1 FROM ancestors
+          WHERE ancestors.uuid IN (:folderUuids)
+        )
+      `,
```
apsantiso: Won't this trigger a folders look up for every favorited folder? 

Correct me if I'm wrong but the CTE is going to be evaluated first for all the favorited folders and the IN :folderUuids expresion is going to be evaluated afterwards

victor-ferro: You're right. I've fixed it by reversing the direction of the walk: a single descent from the deleted folders builds the subtree using `folders_parent_uuid_index`, and each favorite is then checked once against that set (Postgres resolves it as a hashed SubPlan). Cost goes from favorites × depth to subtree size + favorites:

| subtree deleted | before (2k favs) | now    |
|-----------------|------------------|--------|
| 1 folder        | 163 ms           | 2.3 ms |
| 341 folders     | 164 ms           | 3.5 ms |
| 1,365 folders   | 168 ms           | 9.8 ms |
| 21,845 folders  | 171 ms           | 90 ms  |

One point worth mentioning: few favorites with a very large subtree would give worse results. With 200 favorites deleting 21,845 folders it goes from 6.5 ms to 91 ms, with the crossover at around 1,000 folders. I think the change is still better because the new cost is bounded, one-off and proportional to what the user just deleted, whereas the old one grew without a ceiling as favorites piled up and was charged even for trashing an empty folder. But if we care more about bulk deletes, we should revisit it.

src/modules/favorite/favorite.repository.ts
```diff
@@ -81,6 +86,70 @@ export class SequelizeFavoriteRepository implements FavoriteRepository {
...
+      WITH RECURSIVE subtree AS (
+        SELECT fl1.uuid
+        FROM folders fl1
+        WHERE fl1.uuid IN (:folderUuids)
+
+        UNION
+
+        SELECT fl2.uuid
+        FROM folders fl2
+        INNER JOIN subtree ON fl2.parent_uuid = subtree.uuid
```
apsantiso: Remember to put a limit here just in case there's a user with a folder tree that loops infinitely. 

I've seen cases like that before, it's better to just leave a row alive in favorites, since you're already filtering by NOT DELETED, instead of taking the risk of a query going crazy.

apsantiso (commented):
Let's see how it performs

EDIT: this was an approve comment but added a small comment

src/modules/favorite/favorite.repository.ts
victor-ferro: I think we're covered here: the query uses `UNION` instead of `UNION ALL`, so Postgres discards any uuid it has already produced. If the tree loops, the recursion runs out of new rows and stops on its own.

There's also a catch with adding the limit on top: `UNION` dedupes by all the projected columns, so a `depth` column makes every lap look like a new row and we lose the protection we already have. It's one or the other.

Happy to switch to `UNION ALL` + a small limit if you'd rather have the explicit cap, but one thing about the `NOT DELETED` filter: it looks at the item's own state, and trashing a folder only marks that folder, not its descendants. While the folder sits in the trash its children are still `EXISTS`, so a favorite left behind by a truncated walk would keep showing up in the list for that whole window — which is exactly the bug this PR fixes.

apsantiso (approved):

PR #1123, opened 2026-08-05
chore: add NOT NULL constraint to files.uuid

apsantiso:
Adds `NOT NULL` constraint to `files.uuid` in two phases. Any primary key requires having a valid NOT NULL constraint, however, we cannot directly use `SET COLUMN NOT NULL` as it'll lock the table.

  - **Phase 1**: `ADD CONSTRAINT uuid_not_null_chk CHECK (uuid IS NOT NULL) NOT VALID` 
  - **Phase 2**: `VALIDATE CONSTRAINT uuid_not_null_chk`, validates existing rows. Requires only an `SHARE UPDATE EXCLUSIVE` lock, doesn't block concurrent reads/writes.

If phase 2 fails, times out or cancelled, phase 1 stays applied and phase 2 can be retried without rollback.


migrations/20260806085948-step-1-part2-validate-files-uuid-not-null-constraint.js
```diff
@@ -0,0 +1,19 @@
...
+
+// UUID as PK: Phase 2
+// This migration is in another file in case of a failure
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    // Schema updates / vacuums will be locked until the constraint is validated.
+    await queryInterface.sequelize.query(`
+      SET lock_timeout = '2s';
+      ALTER TABLE files VALIDATE CONSTRAINT uuid_not_null_chk;
```
sg-gs, line 10: How is this converting the uuid in the PK?

apsantiso: It's just a small step. We need to guarantee two things before converting a field into a PK: UNIQUE and NOT NULL, otherwhise, we'll lock the table until the entire table is scanned

sg-gs (dismissed):
Running migration @apsantiso

sg-gs (approved):
Migration run @apsantiso

PR #1128, opened 2026-08-06
chore: set not null attr and drop check constraint

apsantiso:


sg-gs (approved):
Migration run @apsantiso

PR #1132, opened 2026-08-06
[_] chore: create not owned uuid index

apsantiso:
The current index is bound to a constraint, we need a free one.

## Changes
- Create a not owned UNIQUE INDEX.


sg-gs (dismissed):
Running migration @apsantiso

sg-gs (approved):
Migration run @apsantiso

PR #1135, opened 2026-08-10
[_] chore: create free id index

apsantiso:
The current index is tied to the PK constraint so a free id index is required so the column retains an index after the PK constraint drop.


sg-gs (approved):
Migration run @apsantiso 
```

== 20260807120000-step-3-part-1-create-files-id-standalone-index: migrating =======
== 20260807120000-step-3-part-1-create-files-id-standalone-index: migrated (1630.651s)

Done in 1631.67s.
```

PR #1136, opened 2026-08-10
[_] chore: swap id and uuid as PK

apsantiso:


sg-gs (approved):
Migration run @apsantiso

PR #1138, opened 2026-08-10
[_] chore: drop duplicated uuid index

apsantiso:
Drop duplicated index on folders. We already have an index tied to a constraint called `folders_uuid_UNIQUE`

image


apsantiso:
Migration executed @sg-gs

sg-gs (approved):

PR #1148, opened 2026-08-20
[PB-6503] add new bucket key hybrid encryption and decryption

TamaraFinogina:
Adds functions for hybrid encryption/decryption of the bucket key, which is what we will use in the hybrid-v2 version instead of mnemonic

### DIFF
--- a/src/modules/user/user.usecase.ts
+++ b/src/modules/user/user.usecase.ts
@@ -348,7 +348,7 @@ export class UserUseCases {
     }
 
     console.info(
-      `(usersReferralsService.redeemUserReferral) ` +
+      '(usersReferralsService.redeemUserReferral) ' +
         `The user '${uuid}' (id: ${userId}) has redeemed a referral: ${type} - ${credit}`,
     );
   }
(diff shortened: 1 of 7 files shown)


src/modules/user/user.usecase.ts
```diff
@@ -571,23 +567,16 @@ export class UserUseCases {
...
-        );
+        await this.asymmetricEncryptionService.reEncryptHybridCiphertext({
+          ciphertextInBase64: encryptionKey,
+          privateKeyInBase64: privateKey,
+          privateKyberKeyInBase64: invite.isHybrid() ? privateKyberKey : null,
+          newPublicKeyInBase64: newPublicKey,
+          newPublicKyberKeyInBase64: invite.isHybrid()
+            ? newPublicKyberKey
+            : null,
+        });
```
jzunigax2, line 579: is this logic not covered by tests in user.usecase.spec.ts ? I mean, since there is no diff there

TamaraFinogina: No, or more like it should be covered by existing tests. It's the same as before, but before it was split into 2 functions: decrypt and then encrypt; now it's one re-encrypt function with the same input as before

apsantiso (approved):
Usually these kind of PRs are approved by @sg-gs, but this only affects sharings with new users so it's not something too critical. 

Make sure to tell QA to check sharings with the previous hybrid method by creating sharings on prod and checking everything works on the feature test branch, etc.

PR #1151, opened 2026-08-31
[PB-6513, PB-6515] Add argon2 enpoints

TamaraFinogina:
Argon2 endpoints: login & password change. No keys in DTOs, no hash & salt encryption

SDK PR: https://github.com/internxt/sdk/pull/439

### DIFF
--- a/migrations/20260831120000-add-argon2-credentials-to-users.js
+++ b/migrations/20260831120000-add-argon2-credentials-to-users.js
@@ -1,0 +1,19 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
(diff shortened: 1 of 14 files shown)


TamaraFinogina:
@jzunigax2 It's all yours :) Definitely missing tests + fixing duplicates. The logic should be similar to what we have currently but simplified - no hash & salt decryption and no key checks. 

P.S. I was not sure about DTO names and locations, so it's -v2 and near the existing ones.

apsantiso:
Marking as blocked so it does not get closed. 

@sg-gs @jzunigax2 what should we do with this? Asking before it gets forgotten

PR #1152, opened 2026-09-02
fix: do not delete file content when replace re-declares the same file id

neuffer:
## What is changed

`FileUseCases.replaceFile` ends by removing the previous content from the storage bucket. The guard on that delete is currently:

```ts
if (oldFileId) {
  await this.network.deleteFile(user, bucket, oldFileId);
}
```

This PR makes it fire only when the content has actually been superseded:

```ts
if (oldFileId && oldFileId !== newFileId) {
```

plus one new test, and one assertion added to an existing test.

## Why

`oldFileId` is the file's contents id as it was before the update. `newFileId` is what the caller sent (`isFileEmpty ? null : newFileData.fileId`). Nothing compares the two.

So a client that calls `PUT /files/{uuid}` to update metadata while re-declaring the contents id the file already has gets this sequence:

1. the row is updated to that same contents id, which changes nothing;
2. that contents id is then deleted from the bucket via `BridgeService.deleteFile`, i.e. `DELETE /buckets/{bucket}/files/{id}`.

The row survives pointing at content that no longer exists. The delete runs after the row update and its failure is only logged, so there is no rollback, and the endpoint answers `200`.

`ReplaceFileDto` accepts an optional `modificationTime`, which is the natural reason a client would call this endpoint without new content. That is how we found it: we were looking at implementing `Utimens` in the Linux client, so that `cp -p` and `rsync -a` stop failing against the drive.

`deleteFile` takes only `(bucket, fileId)`, so it cannot distinguish a superseded object from the current one when the two ids are equal, and we found no reference-counting contract in the code that would make the current behaviour safe.

### Why versioning does not already cover it

The `shouldVersion` branch preserves the old content instead of deleting it, but `isFileVersionable` requires all of: the type is in `VERSIONABLE_FILE_EXTENSIONS` (`pdf`, `docx`, `xlsx`, `csv`); the user's tier has versioning enabled; and the file is under `maxFileSize`. Every other file type takes the deleting path.

## What this PR does NOT fix, stated explicitly

We had this change reviewed by three independent reviewers alongside the client code that motivated it. **They were clear that this guard closes one specific hole and does not make `PUT /files/{uuid}` safe as a metadata endpoint.** We would rather say so here than have it inferred:

1. **The versioned branch is untouched, and is worse than we first thought.** A same-id call on a `pdf`/`docx`/`xlsx`/`csv` never reaches this guard. It runs `applyRetentionPolicy` (which can mark a genuine historical version `DELETED` to make room) and then writes a version row whose `networkFileId` aliases the live file's. **And because `getUserUsedStorage` adds `fileVersionRepository.sumExistingSizesByUser` to the file usage, that spurious version bills the user for the file's full size while storing no new bytes.** A repeated metadata update on a document would inflate quota and evict real history.
2. **A stale contents id defeats this guard.** A client sends the id it has cached. If the server has meanwhile replaced A with B and deleted A, a late request re-declaring A makes this code see `old=B, new=A` - unequal, so it deletes **B**, the live content. Safe metadata updates need an atomic precondition ("update only if the current contents id is X"), which no current endpoint offers.
3. **Same id with a different `size` is still accepted**, which preserves the content while letting the row misreport its size.
4. **A metadata-only call still does upload work**: `enforceMaxUploadFileSize` runs, so a file that is legitimately stored but larger than the user's *current* limit (after a plan change) makes a bare metadata update return `402`; and `addFileReplacementDelta` takes a Redis lock per call even though it correctly writes no usage row for a zero delta.

## What we think the right fix is

**A dedicated metadata route.** All three reviewers reached this independently, and `updateFileMetaData` / `UpdateFileMetaDto` already exists as the natural home - it currently carries only `plainName` and `type`. Accepting `modificationTime` there, doing no quota, usage, versioning, retention or storage work, would resolve every point above at once, and its presence would give clients a capability check they do not currently have.

**We are not proposing that here**, because it is your architectural call and because this guard is worth having on its own. We would be glad to open it if you want it, and we are happy to be told a different shape is preferable.

**The Linux `Utimens` work that led us here is parked**, precisely because of points 1 to 3. We are not asking you to merge this to unblock us.

## Blast radius of this change

Deliberately minimal. Normal replacements are untouched: a real upload sends a new contents id, so `oldFileId !== newFileId` holds and the old content is freed exactly as before. Truncation to zero sends `null`, which also differs. The only behaviour that changes is the equal-ids case, where the current code destroys the content it was asked to keep.

## Evidence

- The new test fails without the guard, on `Expected number of calls: 0 / Received number of calls: 1` against `bridgeService.deleteFile`, and passes with it.
- The assertion added to the empty-file test was verified by control: widening the guard to `oldFileId && newFileId && oldFileId !== newFileId` - the plausible mistake - fails that test and only that test. Without the assertion the same mistake left the suite green.
- `file.usecase.spec.ts` and `file.controller.spec.ts` are green at 192 tests together; eslint reports 0 errors on both changed files.
- **We have not reproduced this against a running server.** The finding is from the source and the unit tests. If your integration environment can confirm it end to end, that seems worth doing before merge.

## Two unrelated things we noticed and deliberately did not touch

- In `applyRetentionPolicy`, when `maxVersions` is `0`, `versionsToKeep.length === maxVersions` is true for the empty array, so `versionsToKeep.at(-1)` returns `undefined`, is pushed into `versionsToDelete`, and `v.id` then throws.
- In `replaceFile`, `fileUuid` is typed `File['fileId']` while being used as a UUID; it looks like it should be `File['uuid']`.

Both look like separate cleanups rather than part of a behavioural fix, so we left them alone.

### DIFF
--- a/src/modules/file/file.usecase.spec.ts
+++ b/src/modules/file/file.usecase.spec.ts
@@ -3201,6 +3201,14 @@ describe('FileUseCases', () => {
         expect(fileRepository.getZeroSizeFilesCountByUser).toHaveBeenCalledWith(
           userMocked.id,
         );
+        // Truncating to empty still supersedes the old content, so it must be
+        // freed. This is the legitimate cleanup nearest to the same-id guard
+        // above, and the one most easily broken by widening it.
+        expect(bridgeService.deleteFile).toHaveBeenCalledWith(
+          userMocked,
+          mockFile.bucket,
+          'old-file-id',
+        );
       });
 
       it('When replacing with empty file and limit is reached, then it should throw', async () => {
--- a/src/modules/file/file.usecase.spec.ts
+++ b/src/modules/file/file.usecase.spec.ts
@@ -2945,6 +2945,38 @@ describe('FileUseCases', () => {
       });
     });
 
+    it('When the new file id matches the current one, then it should not delete the content from the network', async () => {
+      const mockFile = newFile({
+        attributes: {
+          fileId: 'unchanged-file-id',
+          bucket: 'test-bucket',
+          type: 'bin',
+        },
+      });
+      const replaceData = {
+        fileId: 'unchanged-file-id',
+        size: mockFile.size,
+        modificationTime: new Date(),
+      };
+
+      jest.spyOn(fileRepository, 'findByUuid').mockResolvedValue(mockFile);
+      jest.spyOn(fileRepository, 'updateByUuidAndUserId').mockResolvedValue();
+      jest.spyOn(bridgeService, 'deleteFile').mockResolvedValue();
+
+      await service.replaceFile(userMocked, mockFile.uuid, replaceData);
+
+      expect(fileRepository.updateByUuidAndUserId).toHaveBeenCalledWith(
+        mockFile.uuid,
+        userMocked.id,
+        {
+          fileId: replaceData.fileId,
+          size: replaceData.size,
+          modificationTime: replaceData.modificationTime,
+        },
+      );
+      expect(bridgeService.deleteFile).not.toHaveBeenCalled();
+    });
+
     it('When file exists and modificationTime was passed, then it should replace file data with modificationTime', async () => {
       const mockFile = newFile({
         attributes: { fileId: 'old-file-id-string', bucket: 'test-bucket' },
--- a/src/modules/file/file.usecase.ts
+++ b/src/modules/file/file.usecase.ts
@@ -1077,7 +1077,11 @@ export class FileUseCases {
       });
     });
 
-    if (oldFileId) {
+    // Only drop the previous content when it has actually been superseded. A
+    // client may call replace to update metadata (a modification time, say)
+    // while re-declaring the file id it already has; deleting here would
+    // destroy the content the row still points at.
+    if (oldFileId && oldFileId !== newFileId) {
       try {
         await this.network.deleteFile(user, bucket, oldFileId);
       } catch (error) {


jzunigax2:
thanks for raising this @neuffer, I can confirm no client currently hits this behavior as the file is always uploaded beforehand so they hit this endpoint with a fresh file id. 

We should probably guard against this potential bug, however I am not sure this endpoint is the most apppropiate for updating modificationTime which I understand is what you want

neuffer:
Thanks, that confirms it from your side, and it matches what we found reading the code: `FileOverrider` uploads before it calls, so every real caller arrives with a fresh id and the equal-ids branch is unreachable in practice today. We never hit it at runtime either. It is a latent bug rather than a live one, and the PR is written on that basis.

On the endpoint: we agree, and that is our own conclusion too. It is in the PR body under "What we think the right fix is" - three independent reviewers reached it separately, and `updateFileMetaData` / `UpdateFileMetaDto` is the natural home because a metadata route does no quota, usage, versioning, retention or storage work. `PUT /files/{uuid}` cannot be made safe for this by adding a guard, for the reasons in points 1 to 4 of that section, and the guard here does not claim otherwise.

Two separate things, then:

**This PR stands on its own.** It closes one specific hole - a call that re-declares the id the file already has deletes the content it was asked to keep - and it is worth having whether or not anything changes about `modificationTime`. Normal replacements are untouched.

**The metadata-endpoint change is built.** Accepting `modificationTime` on `PUT /files/{uuid}/meta`, with the rename path stopping its overwrite of that field, against `master`, with tests. We held it back because it is your architectural call, and your comment is the answer we were waiting for. Say the word and we will open it; we are equally happy to be told a different shape is preferable before we do.

One thing we would flag either way: the rename path currently stamps `modificationTime = new Date()` on every rename, while `moveFile` never touches the field. Since `modificationTime` is the contents' modification time, a rename should not move it. That is a contract fix independent of accepting the field from a client, and it is the smaller half of what we have.

neuffer:
Following up on the metadata endpoint, and one practical thing first: we can't open the PR. The compare page says *"An owner of this repository has limited the ability to open a pull request to users that are collaborators on this repository."* Something changed after #1152 went up on 2 September, since that one went through the same way. Not a complaint, just so you know why the branch is sitting there rather than arriving as a PR.

It is pushed and ready at `neuffer:feat/meta-accepts-modification-time`, rebased onto current `master` (`4ad8a5e`). Three commits, `npx jest src/modules/file` at 402 from 388 on master, `tsc` clean, eslint unchanged.

Whatever you prefer: pull the branch, cherry-pick it, have someone with access open it, or tell us it is not wanted. Any of those is fine.

On the substance, there are three things in it and we think all three are worth having regardless of who lands them:

1. **A rename should not restamp `modificationTime`.** The field is the contents' modification time and a rename does not change contents. `moveFile` already leaves it alone, so today the two endpoints disagree with each other. This is the smallest and least arguable of the three.
2. **`isNameChanging` currently asks the wrong question.** It compares the resolved name against the stored `plainName` column, so for any row where `plainName` is null the resolved value is the decrypted name and the comparison is string-against-null. Every metadata-only request on such a row looks like a rename, runs the duplicate lookup, and can answer `ConflictException` for a file nobody renamed. Related: `findByPlainNameAndFolderId` takes no uuid to exclude, so a file can conflict with itself.
3. **The update writes `plainName`, `name` and `type` on every request**, including ones carrying only a time, which silently backfills `plainName` on older rows.

We had it reviewed independently before sending, and that review found a compatibility change we had missed and would rather flag than have you discover: `master`'s validation pipe has no `whitelist`, so an unknown `modificationTime` is accepted and ignored today. Declaring it with `@IsDateString()` means a rename carrying `""` or a malformed value would now fail validation and take the rename with it. We transform the empty string away so those keep working; a malformed non-empty value is rejected deliberately. That is in the PR body too.

Happy to split it, reshape it, or drop parts of it.


One last practical note, since the PR route is closed: **this thread is currently the only way we can hand you code.** We can still comment here and we could open an issue, but neither carries a reviewable branch the way a PR does. If a thread on an open PR is an awkward place for this, `hello@internxt.com` works too, or tell us where you would rather have it. We are not trying to route around whatever the restriction is for, we just do not want a fix to go nowhere because there was no channel for it.

neuffer:
Two follow-ups to this thread are ready as branches on my fork, but I see the repository now limits opening pull requests to collaborators, so I cannot send them the usual way. How would you prefer contributions to arrive?

- `neuffer:fix/repair-temporal-file-deleter-merge` - `main` currently fails 9 tests across 3 files. #457 gave `TemporalFileDeleter` a second constructor argument, and #452, merged twenty minutes later, carried tests that construct it with one. Neither pull request was wrong when it was reviewed; each was green against a `main` that did not yet contain the other, and git merged them with no textual conflict. Test files only, and production wiring was never affected. It takes the suite from 9 failing to 1665 passing, and the type-check count from 137 to 135.

- `neuffer:fix/backup-deletes-content-on-unproven-failure` - on the backup path the client deletes the uploaded content whenever the metadata write reports an error. That error does not prove the write failed: an attempt can commit the row and still report failure, and since `INTERNAL_SERVER_ERROR` is retryable the retry then meets the row the first attempt created and comes back as a conflict. Because `deleteFileByFileId` removes the file row when one references that contents id, the single case the client cannot distinguish is exactly the case where a file gets destroyed rather than a stray object.

Happy to send either as a patch, or to leave them if you would rather take a different approach.

One question while I have you, since it changes what the second branch should do. How often does the unreferenced-content cleanup actually run? I found `bin/cleaner/delete-unreferenced-files.ts` in `internxt/bridge`, and `deleteFilesInTheNetworkButNotInDrive()` is exactly the reconciliation this needs, but it reads as an operator-run CLI. If it runs regularly then a leaked object is temporary and the simpler change of never deleting from the client is the better one, and I would happily send that instead. If it is run by hand, then every orphan sits against the user's quota until someone runs it.

apsantiso:
Hi, thank you for raising a potential bug. We'll take it into account for a small fix to prevent requests with the same id in both `newFileId` and `oldFileId`.

However, as my coworker already mentioned, all of our clients generate a new file id before using this endpoint. Any metadata update should be handled by a different endpoint, this one is only intended for replacing content. I'm not sure why would you like to update modificationTime without updating name or content tho.

If you need us to fix a bug, please raise it through our official channels instead of opening a PR trying to force your way in fixing it with an AI agent, one of our collaborators will surely take a look at it and make an approach that fits better taking into account all of the other elements that interact with this repository.

neuffer:
Understood, and thanks for the straight answer - I will use the official
channels from here.

On the modificationTime question, the short answer is that this is not a
nice-to-have: **as the client ships today, a file copied onto the drive loses
its timestamp entirely.** Every file ends up stamped with the moment it was
uploaded rather than the moment it was last written, and that makes the drive
unusable for a whole class of ordinary work.

The tools that hit this are the everyday ones. `cp -p`, `cp -a`, `rsync -t` or
`-a`, `tar -x`, `unzip -o`, `make` and every build system that decides what is
stale by comparing mtimes, backup and archiving tools, photo and music
managers that sort by capture or edit date. None of them are updating a name
or the content. They are preserving a timestamp that already means something,
and on your drive it is silently discarded.

Concretely, `cp -p` sets the time on the OPEN descriptor, before close. From
strace:

    openat(AT_FDCWD, "dst", O_WRONLY|O_CREAT|O_EXCL, 0600) = 4
    utimensat(4, NULL, [...], 0)                           = 0
    close(4)                                               = 0

So the request arrives while the file is still only a local staged copy with no
drive record at all. internxt/drive-desktop-linux#457 handles that case: the time is held and sent with
`POST /files` when the upload creates the file, which is a normal part of
creating one and needs nothing new from you. That merged on 4 September, after
2.6.2, so it is not in any released build yet - which is why the behaviour you
would see today is still "every file carries its upload time".

The case that is still impossible is a file **already on the drive**. The
client refuses `utimens` for those with ENOSYS, deliberately, because the only
endpoint that carries a modificationTime for an existing file is
`PUT /files/{uuid}` - and using it means re-declaring the file's current
contents id purely to change a timestamp. That is exactly the request this PR
was about: on a replace path that treats a re-declared id as superseded, it can
destroy the file's own content, and it mints a billed version for a metadata
change. So the client would rather do nothing than risk that, and the user's
timestamp is lost.

That is the whole reason I care about the guard you have agreed to add. It is
the precondition for ever setting a timestamp on a file that already exists. If
the right answer is a different endpoint - a metadata-only PATCH that takes a
modificationTime and touches nothing else - that would solve it cleanly and I
would be glad to hear that is the plan. I am not asking for it here, just
answering why the question comes up at all.

One thing worth passing to whoever owns `drive-desktop-linux`, unrelated to
any of the above and needing nothing from me: `main` there is currently red,
9 tests across 3 files. internxt/drive-desktop-linux#457 gave `TemporalFileDeleter`
a second constructor argument, and internxt/drive-desktop-linux#452, merged about
twenty minutes later, brought tests that
construct it with one. Neither was wrong when it was reviewed - each was green
against a `main` that did not yet contain the other, and git merged them with
no textual conflict, so nothing flagged it. Your suite runs on `pull_request`
but not on pushes to `main`, which is why it landed silently and why it can
happen again. Whoever picks it up will find it in a minute.

Thanks for the fix on the id guard, and for taking the time to explain.

PR #1154, opened 2026-09-03
[BR-2298]: stabilize listing pagination with a uuid sort tiebreaker

terrerox:
## Problem
A user reported that `rclone ls` / `lsf` / `lsjson` return duplicated filenames and omit other files, while the same folder looks correct in the Drive web app. Reproducible on every listing.

Cause: Folder listings are paged (`limit=50` + `offset`) with no tiebreaker in the sort. Files tied on the sort field (`photo.jpg` / `photo.json` → same `plain_name`; bulk uploads → same `updatedAt`) have no deterministic order, so ties can shift between page queries. When a tie group straddles a page boundary, one file is returned twice and another is never returned.

rclone shows this raw (duplicate + missing files, as reported); the web app hides it by de-duplicating on uuid — but the missing files are absent there too.

## Solution

Append `uuid ASC` as a final sort key in the file and folder repositories when the requested sort has no unique field:

```sql
ORDER BY plain_name COLLATE "custom_numeric" ASC, uuid ASC

Ties become impossible, so pages tile together exactly once per file.

### DIFF
--- a/src/lib/sort.ts
+++ b/src/lib/sort.ts
@@ -1,0 +1,43 @@
+import { Sequelize } from 'sequelize';
+import type { Model, ModelStatic } from 'sequelize';
+import type { Literal } from 'sequelize/types/utils';
+
+const UNIQUE_SORT_FIELDS = new Set(['uuid', 'id']);
+
+/**
+ * Applies the custom numeric collation to plainName sorts and appends a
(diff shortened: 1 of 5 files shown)


sg-gs:
Do not merge this @terrerox. @apsantiso will review and handle it when he comes back.

PR #1166, opened 2026-09-14
[_] fix(recents): show only drive files in recents

apsantiso:
##
Limit the files in /recents to drive files only.

## Why

`/recents` was returning files from any bucket the user owns, including buckets unrelated to the main Drive (e.g. Photos, Backups).


src/modules/file/file.repository.ts
```diff
@@ -909,6 +909,7 @@ export class SequelizeFileRepository implements FileRepository {
         userId,
         status: FileStatus.EXISTS,
         updatedAt: { [Op.gte]: since },
+        ...(bucket && { bucket }),
```
CandelR, line 912: maybe better add tests to new `bucket` in where

CandelR (approved):

PR #1167, opened 2026-09-14
[_] fix/cursor sync microsecond precision

apsantiso:
## What
Preserve full microsecond precision when reading `updated_at` for cursor-based sync pagination, in both files and folders. Probably not something we would see frequently, but it's worth it to guarantee the sync is working correctly (for example, if we add bulk files creation users could end up with a lot of files with the same timestamp, including microseconds)

  - Added `cursor-sql.util.ts`: reads `updated_at` via `to_char()` as a raw string (not JS `Date`), and builds the keyset filter `(updated_at, uuid) > (:cursorTimestamp::timestamptz, :cursorTieBreaker)`.
  - `file.repository.ts` / `folder.repository.ts`: cursor-based sync queries now return `lastRowCursorUpdatedAt` (raw string) instead of the row's `Date`.
  - `file.usecase.ts` / `folder.usecase.ts`: `nextCursor` is encoded from that raw string instead of `updatedAt.toISOString()`.
  - Added e2e coverage for `/files/sync` cursor pagination (no dupes/skips, status filter, invalid/mismatched cursor, empty page).

  ## Why
  Postgres `timestamptz` stores microsecond precision, JS `Date` (and Sequelize's default parsing) truncate to milliseconds. Two rows updated within the same millisecond but different microseconds could get the same cursor, causing duplicate or skipped rows on the next page.


src/common/utils/cursor-sql.util.ts
```diff
@@ -0,0 +1,34 @@
+import { Op, Sequelize, type WhereOptions } from 'sequelize';
+
+// Postgres keeps microseconds, JS Date doesn't — read/compare timestamps as raw strings so we do not lose precision when using keyset pagination
+
+/** Raw SELECT expression preserving full microsecond precision of a timestamptz column. */
```
apsantiso, line 5: I decided to leave comments as it is not 100% clear why this has been done, it's not something you can figure out by just reading the code.

sg-gs (approved):

PR #1169, opened 2026-09-16
[_] fix: remove not optimized brin index and optimize folder creation

apsantiso:
## Changes
- Removed BRIN index, the planner is using it but it fetches almost all the pages, the correlation is too low. We have already the index `idx_deleted_files_updated_at` to cover any query using that index. The correlation is too low to use BRIN here (~0.35)
- Added remove to the WHERE clause when looking for existent folders in POST /folders


sg-gs (approved):
Will execute migration once db PG18 issue is solved @apsantiso

sg-gs (commented):
Migration run @apsantiso 

```

== 20260916120634-drop-brin-indexes-files: migrating =======
== 20260916120634-drop-brin-indexes-files: migrated (425.211s)
```

PR #1172, opened 2026-09-17
[BR-2340]: fix(filters): map current folder unique constraint name to 409

jzunigax2:
## What

`UniqueConstraintFilter` still mapped `folders_plainname_parentid_key`, a constraint that no longer exists. It is now `folders_parentuuid_plainname_unique`, so duplicate folder creations no longer matched any entry and fell through as a **500** instead of a **409**.

- Update the constraint name in the filter's message map.
- Added e2e tests for files and folders that insert a real duplicate row, assert Sequelize raises a `UniqueConstraintError`, and assert the filter maps it to the expected `ConflictException`.

## Why

Production shows `POST /folders` answering 500 with `SequelizeUniqueConstraintError` for folders that already exist, from drive-web, the CLI and rclone's native backend. Clients that retry on a 409 by reusing the existing folder instead see a server error.

The e2e tests bind the filter to the constraint names the database actually has, so the next migration that renames one fails a test rather than silently turning duplicates into 500s again. The files entry (`idx_files_folder_user_name_type_unique_numeric`) needed no change, and the new test confirms it is still current.


apsantiso (dismissed):

jzunigax2:
Added blocked label, actually holding this off until next cli release as per RFC this 409 would get misinterpreted

